PR #27, opened 2023-02-24
Rename addSessionMessage -> sendSessionMessage

petersalas:
#25 missed one of the renames.


juberti (approved):
Does the integration test even work? Seems like this sort of thing should have been caught by it.

petersalas:
> Does the integration test even work? Seems like this sort of thing should have been caught by it.

Nope it doesn't work (the step is marked `continue-on-error: true`) -- that's the thread I started pulling on that got me here.

juberti:
welcome to the team :-)

PR #30, opened 2023-02-28
Add support for OAuth to the SDK

hessamb:
This adds an `OAuthHandler` object to the SDK that agent developers can use to make oauth requests.

This PR also moves `OAuthHandler` and `UserStorage` under a `RunHelper` object that funcs can optionally accept as a second argument.

To make an OAuth agent, one developer needs to first pass their `oauth_params` to the agent constructor (along with `base_prompt` and `fewshots`). The pattern looks as follows:

```python
agent = CodeShotAgent(agent_id, base_prompt, fewshots, oauth_params=oauth_params)

@agent.register_func
def func_that_auths(query, run_helper):
    user_token = run_helper.oauth_handler.user_token()
    if user_token is None:
        return run_helper.oauth_handler.get_authorization_url()
    else:
        client = Client(user_token)
    # do stuff with the authenticated client
```


fixieai/__init__.py
```diff
@@ -3,6 +3,8 @@
 from fixieai.agents import CodeShotAgent
 from fixieai.agents import Embed
 from fixieai.agents import Message
+from fixieai.agents import OAuthParams
+from fixieai.agents import RunHelper
```
juberti, line 7: `RunHelper` isn't super-intuitive to me. Maybe something like `Context` or `Agent` to carry this info?

hessamb: Per our discussion, I changed the API so that Funcs can select what arguments do they want to accept through the argument names.
I think it looks pretty neat :) I think we can support this in most interpreter-based languages.

juberti: It's either this or request.query, request.user_storage, request.oauth_handler. Let's try this and see how it goes.

PR #32, opened 2023-02-28
Return agent config in YAML format

juberti:
Looks a lot nicer when viewing it in the Replit window!

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import functools
 import re
 import threading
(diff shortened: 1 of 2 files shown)


fixieai/agents/code_shot.py
```diff
@@ -152,6 +154,10 @@ def func(query):
 
     def _handshake(self) -> AgentMetadata:
         return AgentMetadata(self.base_prompt, self.few_shots)
```
mdwelsh, line 156: This won't wok with the two return statements

mdwelsh (approved):
LGTM but is there. abug?

fixieai/agents/code_shot.py
juberti: yeah, I fat-fingered the merge, fixing

hessamb (approved):

PR #33, opened 2023-03-02
Reduce the number of python dep coming with fixieai

hessamb:
I think the other deps are used somewhere

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,15 +20,12 @@ click = "^8.1.3"
 fastapi = { version = "^0.89.1", extras = ["all"] }
 gql = { version = "^3.4.0", extras = ["all"] }
 python = "^3.8"
-pillow = "^9.0.1"
 prompt-toolkit = "^3.0.31"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -20,15 +20,12 @@ click = "^8.1.3"
 fastapi = { version = "^0.89.1", extras = ["all"] }
 gql = { version = "^3.4.0", extras = ["all"] }
 python = "^3.8"
-pillow = "^9.0.1"
 prompt-toolkit = "^3.0.31"
 PyYAML = "^6.0"
```
juberti, line 24: do we use this? I import straight up `yaml`

juberti (approved):

pyproject.toml
hessamb: Not sure. Removing it doesn't affect poetry.lock at all. Went ahead and removed this one as well.

mdwelsh (approved):

PR #34, opened 2023-03-02
Authenticate Func calls with query.access_token instead of header

hessamb:
An alternative to fixie-ai/fixie#771


hessamb:
Going with fixie-ai/fixie#771

PR #36, opened 2023-03-03
Use slimmer runtime dependencies

petersalas:
- Make optional `fastapi` dependencies exclusively dev dependencies
- Only include `requests` transport for `gql`

This reduces the installation disk footprint by ~45% (which helps reduce install time accordingly).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -201,7 +201,7 @@ test = ["anyio[trio] (>=3.2.1,<4.0.0)", "black (==22.10.0)", "coverage[toml] (>=
 name = "ghp-import"
 version = "2.1.0"
 description = "Copy your docs directly to the gh-pages branch."
-category = "main"
+category = "dev"
 optional = false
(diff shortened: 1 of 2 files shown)


juberti (approved):
LG but let's add an agent-only super light config

petersalas:
@juberti I played around with poetry dependency groups and extras but ran into an issue with each:
- dependency groups in poetry are explicitly dev only -- they don't/can't appear as dependencies in the built wheel (i.e. `pip install fixieai` would break)
- extras would work but would require that users do `pip install fixieai[client]` or whatever to get the client/cli stuff (and it seems like there's no way to make the default _include_ extras)

I think there are ways to do this that involve publishing multiple packages to PyPi but I'd like to defer that for now.

juberti:
OK, thanks for looking into this, we can revisit once we get the rest of the deploy stuff in place.

PR #39, opened 2023-03-04
Fixie CLI skeleton and init

hessamb:
This moves cli to `fixie/cli` directory and adds support for `fixie init` command. The agent config can be loaded and saved via `load_config()` and `save_config()` from/to `agent.yaml` file.


fixieai/cli/agent_config.py
```diff
@@ -0,0 +1,135 @@
...
+import yaml
+
+from fixieai.cli import validators
+
+
+@dataclasses.dataclass
+class AgentConfig(dataclasses_json.DataClassJsonMixin):
+    """Represents an agent.yaml config file."""
+
+    agent_id: str = "agent_id"
```
mdwelsh, line 16: I'm not sure we want this default, right?

fixieai/cli/agent_config.py
```diff
@@ -0,0 +1,135 @@
...
+            "{cwd}/agent.yaml".
+    """
+    if path is None:
+        path = os.path.join(os.getcwd(), "agent.yaml")
+    with open(path, "w") as fp:
+        fp.write(agent_config.to_yaml())
+
+
+def prompt_user(agent_config: AgentConfig):
+    """Prompts the to edit values in AgentConfig."""
```
mdwelsh, line 62: typo: `Prompts the user to edit values in AgentConfig`

fixieai/cli/cli.py
```diff
@@ -0,0 +1,97 @@
...
+import rich.console as rich_console
+
+import fixieai
+from fixieai.cli import agent_config
+from fixieai.cli import console
+
+textconsole = rich_console.Console()
+
+
+#######  fixie CLI definition  #######
```
mdwelsh, line 13: Rather than this kind of comment I would recommend a module-level docstring comment.

https://google.github.io/styleguide/pyguide.html#382-modules

Doing this will also populate the autogenerated documentation appropriately.

fixieai/cli/cli.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@click.group()
+@click.option("--verbose", is_flag=True, help="Enable verbose output.")
+@click.pass_context
+def fixie(ctx, verbose):
+    """fixie CLI to run and deploy agents."""
```
mdwelsh, line 20: Recommend instead

```
"""Command-line interface to the Fixie platform."""
```

fixieai/cli/cli.py
```diff
@@ -0,0 +1,97 @@
...
+@sessions.command("list", help="List sessions.")
+@click.pass_context
+def list_sessions(ctx):
+    client = ctx.obj["CLIENT"]
+    session_ids = client.get_sessions()
+    for session_id in session_ids:
+        textconsole.print(f"[green]{session_id}[/]")
+
+
+@sessions.command("new", help="Creates a new session and opens it.")
```
mdwelsh, line 78: I still think it would be better to just have `fixie console` instead of `fixie sessions new`.

fixieai/cli/cli.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@sessions.command("open", help="Show session.")
+@click.pass_context
+@click.argument("session_id")
+def open_session(ctx, session_id: str):
+    client = ctx.obj["CLIENT"]
+    session = client.get_session(session_id)
+    messages = session.get_messages()
+    textconsole.print(messages)
```
mdwelsh, line 93: I think this command should be `show`, not `open`, since `open` implies that it will be interactive.

fixieai/cli/validators.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+AGENT_ID_PATTERN = re.compile(r"^\w+$")
+URL_PATTERN = re.compile(
+    r"^https?:\\/\\/(?:www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b"
+    r"(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$"
+)
+
+
+def validate_agent_id(agent_id: str):
```
mdwelsh, line 10: Technically this is an agent `handle`, not an agent ID. For consistency I think we should use the term `handle` here. The `agent_id` is in the format `username/handle`.

fixieai/cli/agent_config.py
```diff
@@ -0,0 +1,135 @@
...
+        agent_config.more_info_url,
+        validators.validate_url,
+    )
+    agent_config.public = _prompt_user_for_bool(
+        "public?",
+        agent_config.public,
+    )
+
+
+def _prompt_user_for_str(
```
mdwelsh, line 90: I believe this code is unnecessary if you use Click's option prompting support.

https://click.palletsprojects.com/en/8.1.x/options/#prompting

The idea here being that the options should be specified on the command line, using flags, but if the flags are not provided then Click prompts the user for them. The resulting value is then passed down as an option in the same way as if it were specified on the command line. This is more consistent and saves code.

fixieai/cli/validators.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+AGENT_ID_PATTERN = re.compile(r"^\w+$")
```
mdwelsh, line 3: It would be better to use the existing `validators` package for this, in this case `validators.slug`. https://validators.readthedocs.io/en/latest/#module-validators.slug

(The agent handle should be a slug so this is the appropriate check.)

fixieai/cli/validators.py
```diff
@@ -0,0 +1,17 @@
...
+URL_PATTERN = re.compile(
```
mdwelsh, line 4: Instead use `validators.url`: https://validators.readthedocs.io/en/latest/#module-validators.url

fixieai/cli/validators.py
```diff
@@ -0,0 +1,17 @@
...
+    r"(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$"
+)
+
+
+def validate_agent_id(agent_id: str):
+    if not AGENT_ID_PATTERN.match(agent_id):
+        raise ValueError("agent_id must be alpha numerics and underscores only.")
+
+
+def validate_url(url: str):
```
mdwelsh, line 15: If you follow my recommendation above to use Click's built-in prompting for options, you can hook up the validators to Click trivially as well:

https://click.palletsprojects.com/en/8.1.x/options/#callbacks-for-validation

fixieai/cli/cli.py
juberti: Seems like a good alias

juberti (commented):
overall LGTM, will defer to Matt for approval

fixieai/cli/agent_config.py
hessamb: Changed it to a default of the current directory name (following the pattern from `npm init`).

fixieai/cli/cli.py
hessamb: Following your suggestion of using more of click, I reorganized the directories to split subcommands into multiple files. I think it looks nicer now.

fixieai/cli/cli.py
hessamb: The goal is to make it interactive (per [fixie CLI doc](https://www.notion.so/fixieai/fixieai-CLI-9c9ea0e3a2184978b84b084bab17176a)). It's not yet interactive but I will make it in a later PR.

fixieai/cli/agent_config.py
hessamb: Great suggestion. We are using more of click now and I think it's in a much better shape.

fixieai/cli/validators.py
hessamb: I was hesitant to install a new package for this, but went ahead and did it. It seems to be lean package.

fixieai/cli/agent/agent_config.py
```diff
@@ -53,3 +59,11 @@ def save_config(agent_config: AgentConfig, path: Optional[str] = None):
         path = os.path.join(os.getcwd(), "agent.yaml")
     with open(path, "w") as fp:
         fp.write(agent_config.to_yaml())
+
+
+def _slugify(s: str) -> str:
```
mdwelsh, line 64: I'd love to see unit tests for new code like this, it's important so that people (including us) don't break things!

mdwelsh (approved):
Nice! Very clean.

fixieai/cli/agent/agent_config.py
hessamb: Done.

PR #41, opened 2023-03-05
Fixie auth command with OAuth

hessamb:
This adds `fixie auth` command to authenticate against the Fixie platform.

The authentication flow follows an OAuth protocol with a localhost as the redirect callback.


fixieai/cli/auth/commands.py
```diff
@@ -0,0 +1,31 @@
+import click
+
+from fixieai.cli.auth import oauth_flow
+from fixieai.cli.auth import user_config
+
+
+def validate_not_authed(ctx, param, force):
```
mdwelsh, line 7: Nits: Please be sure to add tests and docstrings to these functions when you're done.

fixieai/cli/auth/commands.py
```diff
@@ -0,0 +1,31 @@
...
+              callback=validate_not_authed, expose_value=False)
+def auth():
+    fixie_api_token = oauth_flow.oauth_flow()
+    try:
+        config = user_config.load_config()
+    except FileNotFoundError:
+        config = user_config.UserConfig()
+
+    config.fixie_api_key = fixie_api_token
+    user_config.save_config(config)
```
mdwelsh, line 36: I wonder if it would not be simpler to just store the API token in the system keyring, using the Python keyring library:

https://pypi.org/project/keyring/

mdwelsh (approved):
LGTM - but I am definitely not an OAuth expert! Made a couple of suggestions.

fixieai/cli/auth/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Optional
+
+from fixieai.cli import utils
+
+
+@dataclasses.dataclass
+class UserConfig(utils.DataClassYamlMixin):
+    """Represents user config at ~/.config/fixie/config.yaml"""
+
+    fixie_api_key: Optional[str] = None
```
juberti, line 12:
```suggestion
    fixie_access_token: Optional[str] = None
```

fixieai/cli/cli.py
```diff
@@ -1,23 +1,46 @@
...
 import click
 
 import fixieai
+from fixieai import constants
 from fixieai.cli.agent import commands as agent_commands
+from fixieai.cli.auth import commands as auth_commands
 from fixieai.cli.session import commands as session_commands
 
 
+class ContextObject:
```
juberti, line 14: This is kind of a generic name, maybe CliContext? or ClickContext?

fixieai/client/client.py
```diff
@@ -69,13 +70,7 @@ def __init__(
...
     ):
         self._api_url = api_url or constants.FIXIE_API_URL
-        self._api_key = api_key or constants.FIXIE_API_KEY
-        if not self._api_key:
-            raise ValueError(
-                "No Fixie API key provided. Set the FIXIE_API_KEY environment variable "
-                "to your API key, which can be obtained from your profile page on "
-                "https://app.fixie.ai "
-            )
+        self._api_key = api_key or constants.fixie_api_key()
```
juberti, line 73: let's use the right terminology here, this is self._access_token

fixieai/constants.py
```diff
@@ -20,3 +25,28 @@
 
 # The audience for Fixie's query JWTs.
 FIXIE_AGENT_API_AUDIENCE = "https://app.fixie.ai/api"
+
+
+def fixie_api_key() -> str:
```
juberti, line 30: fixie_access_token

fixieai/cli/auth/oauth_flow.py
```diff
@@ -0,0 +1,65 @@
...
+    else:
+        click.echo("Open this link on your browser to continue:")
+        click.echo()
+        click.echo(f"    {url}")
+
+    # Now we wait to receive an oauth code at the redirect uri.
+    code = manager.wait_and_terminate_authorize_code_process()
+    # Now we exchange the code for a fixie access token.
+    manager.init_with_authorize_code(redirect_uri, code)
+    access_token = manager._access_token
```
juberti, line 55: is this the correct API usage, to have to access a private variable?

fixieai/cli/auth/oauth_flow.py
```diff
@@ -0,0 +1,65 @@
...
+from fixieai import constants
+
+# Fixie CLI Client ID.
+CLIENT_ID = "II4FM6ToxVwSKB6DW1r114AKAuSnuZEgYehEBB-5WQA"
+# Scopes to request access for.
+SCOPES = ["fixie-api-key"]
+# The authorization URL that users click on.
+AUTHORIZE_SERVICE = f"{constants.FIXIE_API_URL}/authorize"
+# The token exchange service that we use internally.
+TOKEN_SERVICE = f"{constants.FIXIE_API_URL}/oauth/token"
```
juberti, line 17: I think I would name this URL differently to avoid confusion with our /oauth redirect uri.

juberti (commented):
Overall LG, let's replace `api_key` with `access_token` across this PR though

fixieai/cli/auth/oauth_flow.py
hessamb: The API expects us to use `manager.get`, `manager.post`, etc. for accessing any endpoint that needs that access token. But if we want to get the access_token itself to use elsewhere, I think that's how we should get it.

fixieai/cli/auth/user_config.py
hessamb: Hmm api_key is the terminology we use on the platform (http://app.fixie.ai/profile) and the environment variable `FIXIE_API_KEY`. I think changing it right now will cause more confusion than clarity. The right next step is probably to make it be an api key instead of a full platform access token. Wdyt?

fixieai/cli/auth/commands.py
hessamb: that seems to be suitable for passwords and things that we want the OS to protect I think. Here, we are following the pattern of `gh` and `gcloud`. They store their keys at `~/.config/gh/hosts.yml` and `~/.config/gcloud/access_tokens.db`, respectively.

PR #42, opened 2023-03-06
Add agent creation/update API to SDK

mdwelsh:
This is the first step in enabling the CLI to deploy agents.


fixieai/agents/code_shot.py
```diff
@@ -169,7 +169,7 @@ def func(query):
 
     def _handshake(self) -> fastapi.Response:
         """Returns the agent's metadata in YAML format."""
-        metadata = AgentMetadata(self.base_prompt, self.few_shots)
+        metadata = AgentMetadata(self.base_prompt, self.few_shots)  # type: ignore[call-arg]
```
juberti, line 172: OOC, what are all these ignores for?

fixieai/client/agent.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+        if "agentByHandle" not in result or result["agentByHandle"] is None:
+            raise ValueError(f"Cannot fetch agent metadata for {self._handle}")
+        return result["agentByHandle"]
+
+    def create_agent(
+        self,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+        queries: Optional[List[str]] = None,
```
juberti, line 108: queries should always be pulled from the agent, they should not be written via GQL

fixieai/client/agent.py
```diff
@@ -0,0 +1,238 @@
...
+
+        query = gql(
+            """
+            query getAgentByHandle($handle: String!) {
+                agentByHandle(handle: $handle) {
+                    agentId
+                    handle
+                    name
+                    description
+                    queries
```
juberti, line 88: queries should always be pulled from the agent via GET, I think

fixieai/client/agent.py
```diff
@@ -0,0 +1,238 @@
...
+        assert isinstance(result["createAgent"]["agent"]["agentId"], str)
+        self._metadata = self.get_metadata()
+        return result["createAgent"]["agent"]["agentId"]
+
+    def update_agent(
+        self,
+        new_handle: Optional[str] = None,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+        queries: Optional[List[str]] = None,
```
juberti, line 168: same here

fixieai/agents/code_shot.py
mdwelsh: For whatever reason mypy does not seem to understand that the dataclasses library creates a constructor for the class with these parameters, and complains about this call. I did some Googling and wasn't able to find a solution other than to suppress the error. If you happen to know I'm happy to do something different and get rid of these!

fixieai/client/agent.py
mdwelsh: Ah, OK. I believe right now the GraphQL endpoint allows these to be updated; I will file a bug to remove that capability.

mdwelsh:
PTAL

juberti (approved):

PR #43, opened 2023-03-07
Draft Embed API

juberti:
Wdyt about this approach using data URIs?

### DIFF
--- a/fixieai/agents/api.py
+++ b/fixieai/agents/api.py
@@ -3,10 +3,11 @@ Fixie ecosystem."""
 
 from __future__ import annotations
 
-import requests
 import base64
+import dataclasses
 from typing import Dict, Optional
 
+import requests
 from pydantic import dataclasses as pydantic_dataclasses
 
 
@@ -29,9 +30,9 @@ class Embed:
         return requests.get(self.uri).content
 
     @content.setter
-    def set_content(self, content: bytes):
+    def content(self, content: bytes):
         """Sets the content of the Embed object as a data URI."""
-        self.uri = f"data:{self.content_type};base64,{base64.b64encode(content).decode('utf-8')}"
+        self.uri = f"data:base64,{base64.b64encode(content).decode('utf-8')}"
 
     @property
     def text(self) -> str:
@@ -39,9 +40,9 @@ class Embed:
         return self.content.decode("utf-8")
 
     @text.setter
-    def set_text(self, text: str):
+    def text(self, text: str):
         """Sets the content of the Embed object as a string."""
-        self.set_content(text.encode("utf-8"))
+        self.content = text.encode("utf-8")
 
 
 @pydantic_dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


fixieai/agents/api.py
```diff
@@ -20,6 +22,28 @@ class Embed:
     # upload an Embed object to Fixie and generate a URL.
     uri: str
 
+    @property
+    def content(self) -> bytes:
```
mdwelsh, line 26: Since the other one is 'text', can we use 'bytes' or 'raw_bytes' or something instead? 'content' is kind of generic.

mdwelsh (approved):

fixieai/agents/api.py
```diff
@@ -20,6 +22,28 @@ class Embed:
     # upload an Embed object to Fixie and generate a URL.
     uri: str
```
hessamb, line 23: How would the SDK user upload their bytes to a uri?

fixieai/agents/api.py
juberti: Requests uses .text and .content and it seems to work pretty well (especially since we have .content_type to describe the bytes).

fixieai/agents/api.py
juberti: oops, that was an old comment. Rewritten.

hessamb: I didn't mean the comment, but functionality-wise. As an SDK user, I have an image or a large amount of text that I want to pass to Fixie. Say I have a Func that parses a PDF or gets a text from a custom endpoint.

juberti: just set `.content` and it will flow back to our system as a data URI, at which point we will upload it to GCS

hessamb: oh that's interesting. Then we don't need the sha hashing as well because the uri will remain the same for the same embeds. I think one fallback is that data will be accessible via a public url but not something to worry about right now.

fixieai/agents/api.py
```diff
@@ -16,10 +18,31 @@ class Embed:
...
 
-    # A public URL where the object can be downloaded. The Embed API can be used to
-    # upload an Embed object to Fixie and generate a URL.
+    # A public URL where the object can be downloaded. This can be a data URI.
     uri: str
 
+    @property
+    def content(self) -> bytes:
+        """Retrieves the content for this Embed object."""
+        if self.uri.startswith("data:"):
```
hessamb, line 27: consider using a standard url format like: "base64://bytes" so we can parse with [urlllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse)

hessamb (approved):

fixieai/agents/api.py
juberti: AFAIK `base64` is not an official scheme, but `data` is: https://www.rfc-editor.org/rfc/rfc2397

PR #44, opened 2023-03-07
Remove duplicated mypy config

petersalas:
#42 added a second mypy config which overrode the existing one in `pyproject.toml` -- this removes it and adds the `fixie-examples` exclusion as well as some `ignore[no-any-return]`s for the code that requires it.

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -169,7 +169,7 @@ class CodeShotAgent:
 
     def _handshake(self) -> fastapi.Response:
         """Returns the agent's metadata in YAML format."""
-        metadata = AgentMetadata(self.base_prompt, self.few_shots)  # type: ignore[call-arg]
+        metadata = AgentMetadata(self.base_prompt, self.few_shots)
         yaml_content = yaml.dump(dataclasses.asdict(metadata))
         return fastapi.Response(yaml_content, media_type="application/yaml")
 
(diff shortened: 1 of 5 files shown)


fixieai/agents/code_shot.py
```diff
@@ -169,7 +169,7 @@ def func(query):
 
     def _handshake(self) -> fastapi.Response:
         """Returns the agent's metadata in YAML format."""
-        metadata = AgentMetadata(self.base_prompt, self.few_shots)  # type: ignore[call-arg]
+        metadata = AgentMetadata(self.base_prompt, self.few_shots)
```
petersalas, line 172: This `ignore[call-arg]` is not necessary with the pydantic mypy plugin

fixieai/client/agent.py
```diff
@@ -44,49 +44,49 @@ def agent_id(self) -> Optional[str]:
         """Return the agentId for this Agent."""
         if self._metadata is None:
             return None
-        return self._metadata["agentId"]
+        return self._metadata["agentId"]  # type: ignore[no-any-return]
```
hessamb, line 47: consider doing an assertion on the dtype instead for these
`assert isinstance(self._metadata["agentId"], str)

hessamb (approved):
LGTM. Thanks for the improvements

juberti (approved):

PR #46, opened 2023-03-08
Allow dev fixie audience, use agent id from token claims

petersalas:
- Use `"aid"` claim from JWT instead of `self.agent_id` for OAuth operations
- Allow fixie dev audience for JWTs
- Add `crypto` extra for `pyjwt` (required for Ed25519)

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -24,6 +24,16 @@ from fixieai.agents import utils
 # Regex that controls what Func names are allowed.
 ACCEPTED_FUNC_NAMES = re.compile(r"^\w+$")
 
+# The JWT claim containing the agent ID
+_AGENT_ID_JWT_CLAIM = "aid"
+
+
+@pydantic_dataclasses.dataclass
+class VerifiedTokenClaims:
+    """Verified claims from an agent token."""
+
+    agent_id: str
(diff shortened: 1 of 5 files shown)


hessamb (approved):

fixieai/agents/code_shot.py
```diff
@@ -211,32 +222,45 @@ def _serve_func(
...
+            claims = jwt.decode(
                 token,
                 public_key.key,
                 algorithms=["EdDSA"],
-                audience=constants.FIXIE_AGENT_API_AUDIENCE,
+                audience=constants.FIXIE_AGENT_API_AUDIENCES,
             )
-            return True
+
+            if _AGENT_ID_JWT_CLAIM not in claims or not isinstance(
```
hessamb, line 235: nit: I would put this logic into a `VerifiedTokenClaims.from_jwt(token, public_key)` that raises `jwt.DecodeError` if bad format. And probably move that dataclass to further down in this file (since this is in SDK and the order will become the order in the docs).

petersalas: I ended up consolidating all token validation into `_VerifiedTokenClaims` and passing the client in to skirt around the awkwardness of raising `jwt.DecodeError`

PR #48, opened 2023-03-08
Remove agent_id from `CodeShotAgent` constructor

petersalas:
- Remove `agent_id` from `CodeShotAgent` constructor
- Make it an optional parameter to `serve` to support unauthed scenarios
- Add `FixieClient.get_current_username()` and `FixieClient.refresh_agent(handle)`
- Add `fixie agent refresh [handle]` CLI command which invokes `refresh_agent` (`handle` is required, but can be inferred from local `agent.yaml` if present)
- Enable mypy `check_untyped_defs` and fix a test violation that was added in #46 by refactoring fixture
- Opportunistically take advantage of fixture refactor to add new test to exercise invalid token scenario

A followup change will update `fixie/fixie-examples` accordingly.

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -5,10 +5,12 @@ import functools
 import inspect
 import json
 import re
+import threading
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 import fastapi
 import jwt
+import requests
 import uvicorn
 import yaml
 from pydantic import dataclasses as pydantic_dataclasses
@@ -107,17 +109,24 @@ class CodeShotAgent:
             # Register default Funcs.
             self.register_func(_oauth)
 
-    def serve(self, host: str = "0.0.0.0", port: int = 8181):
+    def serve(
+        self, agent_id: Optional[str] = None, host: str = "0.0.0.0", port: int = 8181
+    ):
         """Starts serving the current agent at `{host}:{port}` via uvicorn.
 
-        This pings Fixie upon startup to fetch the latest prompt and fewshots.
+        If agent_id is specified, this pings Fixie upon startup to fetch the latest prompt and fewshots.
 
         Args:
+            agent_id: The qualified agent id (`username/handle`)
             host: The address to start listening at.
             port: The port number to start listening at.
         """
         fast_api = fastapi.FastAPI()
         fast_api.include_router(self.api_router())
+        if agent_id:
+            fast_api.add_event_handler(
+                "startup", functools.partial(_ping_fixie_async, agent_id)
+            )
         uvicorn.run(fast_api, host=host, port=port)
 
     def api_router(self) -> fastapi.APIRouter:
@@ -296,3 +305,14 @@ def _split_few_shots(few_shots: str) -> List[str]:
         "Q:" + few_shot for few_shot in few_shot_splits[1:]
     ]
     return few_shot_splits
+
+
+def _ping_fixie_async(agent_id: str):
+    """Asynchronously pings Fixie to refresh the given agent_id."""
+    thread = threading.Thread(target=_ping_fixie_sync, args=(agent_id,))
+    thread.start()
+
+
+def _ping_fixie_sync(agent_id: str):
+    response = requests.post(f"{constants.FIXIE_REFRESH_URL}/{agent_id}")
+    response.raise_for_status()
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -5,12 +5,10 @@ import functools
 import inspect
 import json
 import re
-import threading
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 import fastapi
 import jwt
-import requests
 import uvicorn
 import yaml
 from pydantic import dataclasses as pydantic_dataclasses
@@ -92,7 +90,6 @@ class CodeShotAgent:
 
     def __init__(
         self,
-        agent_id: str,
         base_prompt: str,
         few_shots: Union[str, List[str]],
         oauth_params: Optional[oauth.OAuthParams] = None,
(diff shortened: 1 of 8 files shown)


fixieai/agents/code_shot.py
```diff
@@ -122,9 +118,6 @@ def serve(self, host: str = "0.0.0.0", port: int = 8181):
         """
         fast_api = fastapi.FastAPI()
         fast_api.include_router(self.api_router())
-        fast_api.add_event_handler(
```
hessamb, line 125: Hmm this is going to make the replit path, who currently rely on `.serve()` to refersh their agent a bit less convenient. Example: [this stock agent](https://replit.com/@fixie-ai/Stock#main.py).

One option would be to say, replit agents should run `fixie agent refresh agent_id` after every re-run of their agent.

Another option would be to make agent_id an argument of `.serve(agent_id)` and keep the behavior as is. For the `fixie deploy` path, we expect `main.py` not to include the `.serve()` call. So we can pass the agent_id from `AgentConfig` to `.serve()` when we're adding that as part of `fixie deploy`.

Between these 2, I tend to prefer option 2. Unless there's some other option or argument that I'm missing.

fixieai/agents/code_shot_test.py
```diff
@@ -24,10 +24,18 @@
 """
 
 
+@pytest.fixture(autouse=True)
+def mock_token_verifier():
```
hessamb, line 28: nit: The pytest friendly way of mocking is to use the mocker fixture:
```python
def mock_token_verifier(mocker):
    return mocker.patch(
        "fixieai.agents.code_shot._VerifiedTokenClaims.from_token",
        return_value=code_shot._VerifiedTokenClaims(agent_id="fake_agent_id")
    )
```

fixieai/client/client.py
```diff
@@ -166,3 +168,28 @@ def create_session(self) -> Session:
...
+        user = result["user"]
+        assert "username" in user and isinstance(user["username"], str)
+        return user["username"]
+
+    def refresh_agent(self, agent_handle: str):
+        """Indicates that an agent's prompts should be refreshed."""
+        username = self.get_current_username()
+        requests.post(
+            f"{self._api_url}/api/refresh/{username}/{agent_handle}",
+            headers=self._request_headers,
```
hessamb, line 194: FYI `/refresh` currently can be hit unauthed. Something we can fix if we put refresh as part of cli, but maybe not if we keep it as part of `.serve()`.

fixieai/client/client.py
```diff
@@ -166,3 +168,28 @@ def create_session(self) -> Session:
...
+        assert "user" in result and isinstance(result["user"], dict)
+        user = result["user"]
+        assert "username" in user and isinstance(user["username"], str)
+        return user["username"]
+
+    def refresh_agent(self, agent_handle: str):
+        """Indicates that an agent's prompts should be refreshed."""
+        username = self.get_current_username()
+        requests.post(
+            f"{self._api_url}/api/refresh/{username}/{agent_handle}",
```
hessamb, line 193: optional nit: I tend to think we should get all paths from `constants.py` instead of constructing them here. This `FixieClient` probably should not get the `api_url` in the constructor and follow the pattern elsewhere that all paths come directly from `constants.py`.

fixieai/agents/code_shot.py
juberti: yeah, I tend to agree. The problem right now is that the replit way is deprecated but the new way doesn't work yet, I think we should keep replits working until at least after the hackathon.

fixieai/client/client.py
hessamb: Some more context: We have this [`UrlHub`](https://github.com/fixie-ai/fixie/blob/e869647a84604385f9ba4225e9cf47db8049dcb3/src/backend/fixie/chain/api/agent.py#L109) object on the platform that we intend to pass to the 3P agents that introduces all service urls at some point. Not too relevant here but just FYI.

fixieai/agents/code_shot.py
hessamb: Oops sorry. I just added you to the replit team. Try again please.

Yes that's how I think about it. If the user calls `.serve()` (they are deploying locally), they would need to provide "username/agent_id". If they are deploying to fixie via cli, they don't type in `.serve()` at all and the cli will add that as part of deploy process.

fixieai/client/client.py
petersalas: Ah, yes I agree it's confusing that some places might be using different API URLs -- I removed the argument from the constructor to align this across the SDK.

fixieai/agents/code_shot_test.py
petersalas: Ah, I didn't realize we had `pytest-mock`, thank you!

In concert with removing the api URL from `FixieClient` I also normalized the `requests_mock` usage.

fixieai/agents/code_shot.py
petersalas: I was bothered by being unable to update `examples/agents/randint.py` with an appropriate id so I ended up making this an _optional_ argument for these unauthed/single-file scenarios (i.e. replit). Does that seem reasonable?

fixieai/client/client.py
petersalas: This is true -- I'm inclined to leave the auth headers here because it feels sort of weird that it's not authed (i.e. eventually we might be able to make it authed)? But I can remove it if that's just confusing.

fixieai/agents/code_shot.py
hessamb: Yes that's reasonable.

fixieai/client/client.py
hessamb: Agreed. Let's keep it in.

hessamb (approved):
LGTM!

PR #50, opened 2023-03-08
[cli] Implement fixie serve/deploy commands

petersalas:
This implements `fixie agent serve` and `fixie agent deploy` commands (with a stub for pushing the actual deployment to resolve the chicken/egg problem of using the loader within the deployment itself).

Some notes:
- `AgentConfig.entry_point` is redefined for consistency with `uvicorn` such that `$MODULE:$ATTRIBUTE` points to an agent implementation (default is now `main:agent` instead of `main.py`).
- Define a function that normalizes `Optional[str]` into a path to an `agent.yaml` file: `None` resolves to `./agent.yaml`, paths to directories resolve to `$path/agent.yaml`, otherwise strings are taken as-is.
- Make `AgentConfig.deployment_url` explicitly optional -- when specified, `fixie agent deploy` updates the agent configuration and triggers a refresh. When _not_ specified, `fixie agent deploy` will deploy the agent as a Fixie-managed deployment.
- Add support for updating `funcUrl` from `Agent`
- Remove `fixie agent refresh` as it no longer has value distinct from `fixie agent deploy`


mdwelsh (approved):
Nice!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "fixieai"
-version = "0.1.10"
+version = "0.1.11"
```
hessamb, line 3: I'd not publish this to pypi yet until the next PR

hessamb (approved):
Nice and clean! Looking forward to the follow up

PR #52, opened 2023-03-09
[cli] Implement deployment upload

petersalas:
Implement the actual deployment for `fixie agent deploy`. Consists of:
- Walking the agent subtree to collect every (non-hidden) file in the subtree
- Defining some bootstrapping code at the location the service expects (`main.py`) to listen on the port the service expects (`os.getenv("PORT")`)
- A spinner to show that something is happening during the deployment

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -184,18 +184,22 @@ def deploy(ctx, path):
         # Deploy the agent to fixie with some bootstrapping code.
         file_streams: Dict[str, BinaryIO] = {}
         deploy_root = os.path.dirname(path)
-        for root, dirs, files in os.walk(deploy_root):
-            # Exclude any hidden files
+        for root, dirs, files in os.walk(deploy_root, topdown=True):
+            # Exclude any hidden files by removing them from the dirs list.
+            # As per https://docs.python.org/3/library/os.html#os.walk
+            # removing them from the dirs list with topdown=True skips
+            # them from the walk.
             for i, dir in reversed(list(enumerate(dirs))):
-                if dir and dir[0] == ".":
+                if dir.startswith("."):
                     console.print(
                         f"Ignoring hidden directory {os.path.join(root, dir)}",
                         style="grey53",
                     )
+                    # N.B. We're iterating in reverse to avoid needing to adjust i.
                     del dirs[i]
 
             for filename in files:
-                if filename and filename[0] == ".":
+                if filename.startswith("."):
                     console.print(
                         f"Ignoring hidden file {os.path.join(root, filename)}",
                         style="grey53",
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -1,6 +1,10 @@
+import io
 import os
+from contextlib import contextmanager
+from typing import BinaryIO, Dict
 
 import click
+import rich.console as rich_console
 import validators
 
 import fixieai.client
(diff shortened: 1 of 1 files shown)


mdwelsh (approved):

fixieai/cli/agent/commands.py
```diff
@@ -145,14 +149,72 @@ def serve(ctx, path, host, port):
...
+        deploy_root = os.path.dirname(path)
+        for root, dirs, files in os.walk(deploy_root):
+            # Exclude any hidden files
+            for i, dir in reversed(list(enumerate(dirs))):
+                if dir and dir[0] == ".":
+                    console.print(
+                        f"Ignoring hidden directory {os.path.join(root, dir)}",
+                        style="grey53",
+                    )
+                    del dirs[i]
```
hessamb, line 195: what does this `del` line do?

fixieai/cli/agent/commands.py
```diff
@@ -145,14 +149,72 @@ def serve(ctx, path, host, port):
...
+
+            for filename in files:
+                if filename and filename[0] == ".":
```
hessamb, line 198: nit 
```suggestion
                if filename.startwith("."):
```

fixieai/cli/agent/commands.py
```diff
@@ -145,14 +149,72 @@ def serve(ctx, path, host, port):
...
+
     if config.deployment_url is None:
-        raise NotImplementedError("Managed deployment is not yet implemented")
+        # Deploy the agent to fixie with some bootstrapping code.
+        file_streams: Dict[str, BinaryIO] = {}
+        deploy_root = os.path.dirname(path)
+        for root, dirs, files in os.walk(deploy_root):
+            # Exclude any hidden files
+            for i, dir in reversed(list(enumerate(dirs))):
+                if dir and dir[0] == ".":
```
hessamb, line 190:
```suggestion
                if dir.startwith("."):
```

fixieai/cli/agent/commands.py
```diff
@@ -145,14 +149,72 @@ def serve(ctx, path, host, port):
...
+
+                full_path = os.path.join(root, filename)
+                upload_path = os.path.join(
+                    "agent", os.path.relpath(full_path, deploy_root)
+                )
+                file_streams[upload_path] = open(full_path, "rb")
+
+        file_streams["main.py"] = io.BytesIO(_DEPLOYMENT_BOOTSTRAP_SOURCE.encode())
+        with _spinner(console, "Deploying..."):
+            ctx.obj.client.deploy_agent(
```
hessamb, line 213: We should get back a deployment_url to put in `agent.yaml` right? Or are we not going to expose that to agent developers?

fixieai/cli/agent/commands.py
```diff
@@ -145,14 +149,72 @@ def serve(ctx, path, host, port):
...
+                config.handle,
+                file_streams,
+            )
 
     # Trigger a refresh with the updated deployment
-    ctx.obj.client.refresh_agent(config.handle)
+    with _spinner(console, "Refreshing..."):
```
hessamb, line 219: Does the deployment endpoint on fixie stalls until the deployment is finished and running? Because otherwise hitting refresh before that doesn't yield good results.

fixieai/cli/agent/commands.py
petersalas: Yeah the current model with deployment API is that managed deployments are mutually exclusive with API-visible `funcURL`s (ideally/eventually the deployment would/will be an API entity of its own and then it might be represented as a `String | Deployment` in GraphQL or something like that).

So with this implementation when `agent.yaml` has a `deployment_url` that's interpreted as the user saying "I'll deploy this myself". Which, for the managed deployment scenario means there's no need to write anything back to the YAML file (which IMO is probably a good thing as it makes it more amenable to source control/templates/etc.).

fixieai/cli/agent/commands.py
petersalas: Good callout -- the deployment API certainly waits for the rollout to be successful before returning. (And will fail if it's not.) In my experience so far the next request hits next the newly deployed revision, but this is something I want to test more throughly. (If it's not reliably the case, I'll look make changes to the service API to wait longer rather than changes in the SDK.)

fixieai/cli/agent/commands.py
hessamb: that's reasonable.

fixieai/cli/agent/commands.py
hessamb: sounds good. The unicorn service needs to be up and running before hitting `/refresh`, otherwise, the internal refresh would only result in a HttpError. Another option would be to format the `_DEPLOYMENT_BOOTSTRAP_SOURCE` above to pass the agent_id to the `.serve()` call so the refresh gets triggered by the deployed agent itself. That's pretty reliable.

petersalas: The catch that led to me do it this way is that on the first deployment the URL isn't available (at least I haven't seen a way to get it) until the deployment completes -- `POST /api/deployments` only sets the URL on the agent after the deployment completes.

hessamb: Makes sense. Sorry for dragging here a bit. If GCR deployment finishes after the service is actually up it should be very safe.

hessamb (approved):

fixieai/cli/agent/commands.py
petersalas: Not at all, these are all good questions!

PR #53, opened 2023-03-09
Add CLI support for showing and deleting agents.

mdwelsh:


fixieai/cli/agent/commands.py
```diff
@@ -104,6 +106,38 @@ def list_agents(ctx, verbose):
...
+        click.echo(f"Agent {agent_id} not found.")
+        return
+    click.secho(f"{agent.agent_id}", fg="green", nl=False)
+    click.echo(f": {agent.name}")
+    click.echo(f"Owner: {agent.owner}")
+    click.echo(f"Description: {agent.description}")
+    click.echo(f"More info URL: {agent.more_info_url}")
+    click.echo(f"Published: {agent.published}")
+    click.echo(f"Func URL: {agent.func_url}")
+    click.echo(f"Query URL: {agent.query_url}")
```
juberti, line 125: should we hide fields that are None? I hate to expose system innards unnecessarily.

fixieai/client/agent.py
```diff
@@ -18,44 +18,50 @@ class Agent:
...
-        return self._handle
-
     @property
     def agent_id(self) -> Optional[str]:
         """Return the agentId for this Agent."""
         if self._metadata is None:
             return None
         agent_id = self._metadata["agentId"]
-        assert isinstance(agent_id, str)
+        assert agent_id is None or isinstance(agent_id, str)
```
juberti, line 50: when would this be None? same for the fields below, shouldn't they be "" at worst?

fixieai/client/agent.py
```diff
@@ -100,80 +108,120 @@ def owner(self) -> Optional[str]:
...
+            return datetime.datetime.fromisoformat(ts)
+        else:
+            return None
 
     def get_metadata(self) -> Dict[str, Any]:
         """Return metadata about this Agent."""
 
-        query = gql(
+        if self._owner is None:
+            # Query by handle.
```
juberti, line 158: I assume this scopes to your own user id, rather than returning everything with that handle?

mdwelsh: yes

fixieai/client/agent.py
mdwelsh: Because the agent might not exist yet

fixieai/cli/agent/commands.py
mdwelsh: If you don't mind I'd like to fix this later, trying to get this done quickly and bouncing between a lot of things...

juberti (approved):

PR #62, opened 2023-03-10
minimal docs update

stefania11:
just editing landing page on docs for now


pyproject.toml
```diff
@@ -30,6 +30,7 @@ rich = "^12.6.0"
 dataclasses-json = "^0.5.7"
 validators = "^0.20.0"
 oauth2-client = "^1.3.0"
+cinder = "^21.1.0"
```
mdwelsh, line 33: mkdocs-cinder is already in the dev dependencies. It should not be needed by users of the SDK.

mdwelsh (approved):
LGTM but please remove cinder from the SDK dependencies, since I don't think it is needed.

mdwelsh:
Note that poetry.lock needs to be checked into the source tree as it defines the actual dependencies used by a given checkout of the tree.

docs/index.md
```diff
@@ -22,20 +22,20 @@ Install the Fixie SDK using pip:
 $ pip install fixieai
 ```
 
-Set your `FIXIE_API_KEY` environment variable to your API key, which
-you can find in the Fixie web interface:
-```shell
-$ export FIXIE_API_KEY=<YOUR API KEY>
-```
-
+Once installed, run `fixie auth` to authenticate with Fixie and start building.
```
zkoch, line 25: I think we learned we need to tell them to make sure they've first authenticated and created an account on Fixie. Then they can auth

stefania11:
Cinder is needed when you want to run `just serve-docs` and check the docs
locally

PR #63, opened 2023-03-10
Allow setting frontend-agent-id

juberti:


Justfile
```diff
@@ -70,3 +70,7 @@ build-docs:
 # Deploy docs to Github Pages.
 publish-docs:
     poetry run mkdocs gh-deploy --force
+
+# Run the fixie CLI.
+cli *FLAGS:
```
hessamb, line 75: No need for this. We can do `poetry run fixie ...` in this repo and it picks the latest.

fixieai/cli/session/commands.py
```diff
@@ -18,11 +18,12 @@ def list_sessions(ctx):
 
 
 @session.command("new", help="Creates a new session and opens it.")
+@click.option("-a", "--agent", required=False)
```
hessamb, line 21: Please add a callback that verifies agent exists and exists gracefully if not. See [auth_force_callback as an example](https://github.com/fixie-ai/fixie-sdk/blob/6691d392c500081d8b46ea24a8ed5fe71e5d022a/fixieai/cli/auth/commands.py#L8).

hessamb (approved):
Nice PR! couple of nits

Justfile
juberti: how about I make this `just fixie` which does what you say above?

hessamb: sure that'd be easier to type

fixieai/cli/session/commands.py
juberti: ok, will give it a shot

PR #66, opened 2023-03-10
Add --reload option (enabled by default) to `fixie agent serve`

petersalas:
Add `--reload` option to `fixie agent serve` to automatically reload on agent implementation changes.

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -4,6 +4,7 @@ import dataclasses
 import functools
 import inspect
 import json
+import os
 import re
 import threading
 import time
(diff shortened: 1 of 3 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -241,7 +245,21 @@ def serve(ctx, path, host, port, tunnel):
...
+                host=host,
+                port=port,
+                factory=True,
+                reload=True,
+                reload_dirs=[os.path.dirname(path)],
+                app_dir=os.path.dirname(path),
+            )
+        else:
+            _, agent_impl = loader.load_agent_from_path(path)
+            agent_impl.serve(agent_api.agent_id, host, port)
```
petersalas, line 262: Maybe get rid of `load_agent_from_path`(/`serve`?) altogether and just use `uvicorn` directly in both deploy/serve scenarios.

fixieai/agents/code_shot.py
```diff
@@ -119,20 +120,33 @@ def serve(
...
+
+        If agent_id is specified as an argument or via the FIXIE_REFRESH_AGENT_ID environment variable,
+        this pings Fixie upon startup to fetch the latest prompt and fewshots.
+
+        Args:
+            agent_id: The qualified agent id (`username/handle`)
+        """
         fast_api = fastapi.FastAPI()
         fast_api.include_router(self.api_router())
+        agent_id = agent_id or os.getenv("FIXIE_REFRESH_AGENT_ID")
```
mdwelsh, line 144: Why do you need the env var?

mdwelsh (approved):

fixieai/cli/agent/commands.py
petersalas: Thinking about this more, I think this is worth doing eventually (probably after hackathon), but to minimize risk of incompatible SDK versions during testing or otherwise I'll wait for this change to roll out first.

fixieai/cli/agent/commands.py
```diff
@@ -223,12 +224,15 @@ def _ensure_agent_updated(
     "--tunnel/--no-tunnel",
     is_flag=True,
     default=True,
-    help="Create a tunnel using localhost.run",
+    help="Create a tunnel using localhost.run.",
+)
+@click.option(
+    "--reload/--no-reload", is_flag=True, default=True, help="Enable auto-reload."
```
hessamb, line 230: Can you add "default set" to this flag and the one above please? Right now the help looks a bit confusing and suggests they are not set.

Screen Shot 2023-03-10 at 9 54 53 AM

hessamb (approved):

fixieai/cli/agent/commands.py
petersalas: Good callout! Will do

PR #72, opened 2023-03-10
Use `fixie console --agent` instead of @ syntax

petersalas:


fixieai/cli/agent/commands.py
```diff
@@ -352,7 +352,9 @@ def deploy(ctx, path, metadata_only):
     else:
         suggested_query = "Hello!"
 
-    suggested_message = shlex.quote(f"@{agent_api.agent_id} {suggested_query}")
+    suggested_command = (
+        f"fixie console -a {agent_api.agent_id} {shlex.quote(suggested_query)}"
```
hessamb, line 356: tbh I'm not a big fan of short flags when they're not popular yet
```suggestion
        f"fixie console --agent {agent_api.agent_id} {shlex.quote(suggested_query)}"
```

hessamb (approved):

PR #76, opened 2023-03-10
Escape brackets for textconsole formatting

hessamb:
Rich console takes everything inside `[ ... ]` as its own style object. This brings back func_name in `Ask Func[func_name]` to console prints.

### DIFF
--- a/fixieai/cli/session/console.py
+++ b/fixieai/cli/session/console.py
@@ -5,9 +5,9 @@ from typing import Any, Dict, List, Optional
 import prompt_toolkit
 import prompt_toolkit.history
 import requests
+from PIL import Image
 from rich import console as rich_console
 from rich import markup
-from PIL import Image
 
 from fixieai import FixieClient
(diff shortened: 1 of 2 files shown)


fixieai/cli/session/console.py
```diff
@@ -58,13 +59,13 @@ def run(
...
 
     def _query(self, in_text: str) -> None:
-        with textconsole.status("Working...", spinner="bouncingBall"):
-            try:
-                for message in self._session.run(in_text):
-                    self._show_message(message)
-            except requests.exceptions.HTTPError as e:
-                textconsole.print(f"🚨 {e}")
-                return
+        # with textconsole.status("Working...", spinner="bouncingBall"):
```
petersalas, line 62: leftover from debugging?

petersalas (approved):

fixieai/cli/session/console.py
hessamb: woops

PR #77, opened 2023-03-10
Change to agent directory in `fixie serve`/`fixie deploy`

petersalas:
This ensures that path resolution is consistent between the two (and independent of the CLI-invoker's working directory).


hessamb (approved):
Would you bump the SDK version and publish after merge please

PR #83, opened 2023-03-11
Relax prompttoolkit version, and convert windows path to linux on upload

hessamb:


pyproject.toml
```diff
@@ -22,7 +22,7 @@ fastapi = { version = "^0.89.1" }
 uvicorn = { version = "^0.20.0", extras = ["standard"] }
 gql = { version = "^3.4.0", extras = ["requests"] }
 python = "^3.8"
-prompt-toolkit = "^3.0.31"
+prompt-toolkit = "*"
```
mdwelsh, line 25: This is kind of dodgy, is this the only way to do this?

mdwelsh (approved):

petersalas (approved):

pyproject.toml
hessamb: This was causing a conflict with colab. Colab expects a range (2.x.x to 2.xx). By default, pip resolves to the latest version that satisfies the constraints.

PR #86, opened 2023-03-13
Generate `requirements.txt` from `fixie init`

petersalas:
This adds support for generating `requirements.txt` via `fixie init`. The command now takes an optional `--requirement` argument (which can be specified multiple times) to include requirements in a `requirements.txt` file.

Even if no requirements are specified, the command will include a `fixieai` requirement specifier that is version-compatible with the user's SDK. (To facilitate this, this also adds `__version__` to the top-level `__init__.py`.)

If the file already exists, the arguments are taken as additions unless they are already present. If an existing `fixieai` requirement is specified that particular requirement will be _replaced_.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -1,4 +1,5 @@
 import contextlib
+import functools
 import io
 import os
 import pathlib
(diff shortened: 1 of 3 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -63,6 +68,39 @@ def _validate_entry_point(ctx, param, value):
             return value
 
 
+def _update_agent_requirements(
```
mdwelsh, line 71: Nit: I'd have a unit test for this stuff since it's a little complex.

mdwelsh (approved):

fixieai/cli/agent/commands.py
petersalas: Yes! Good call.

PR #87, opened 2023-03-14
Make YAML formatting more natural to match existing samples

petersalas:
This changes `agent.yaml` formatting to:

- Exclude `None`-valued fields whose defaults are also `None` (namely `name` and `deployment_url`)
- Serialize strings using `|` if and only if they are multiline
- Reorder `more_info_url` and `entry_point`

With these change `fixie init` produces YAML files similar to those in the fixie-examples repo.


mdwelsh (approved):

juberti (approved):
Nice!

PR #91, opened 2023-03-18
Add agent validation to serve/deploy

petersalas:
This adds stricter agent validation to `fixie serve` and `fixie deploy`. In particular:
- Agents now run from within their own virtual environment in which dependencies from `requirements.txt` are installed
- `python -m fixieai.cli.agent.loader` now loads and validates the agent

This also moves fewshot validation out of the agent constructor and into `load_agent_from_path` so that it has the opportunity to run _after_ all funcs have been registered.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -6,20 +6,21 @@ import pathlib
 import random
 import re
 import shlex
+import subprocess
+import sys
 import urllib.request
+import venv
 from contextlib import contextmanager
-from typing import BinaryIO, Dict, List
+from typing import BinaryIO, Dict, List, Tuple
 
 import click
 import rich.console as rich_console
-import uvicorn
 import validators
 
 import fixieai.client
 from fixieai import constants
 from fixieai.cli.agent import agent_config
-from fixieai.cli.agent import loader
-from fixieai.cli.agent import tunnel as tunnel_
+from fixieai.cli.agent import tunnel
 
 # Regex pattern to match valid entry points: "module:object"
 VAR_NAME_RE = r"(?![0-9])\w+"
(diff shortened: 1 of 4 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -392,10 +470,26 @@ def _spinner(console: rich_console.Console, text: str):
     is_flag=True,
     help="Only publish metadata and refresh, do not redeploy.",
 )
+@click.option(
+    "--validate/--no-validate",
+    "validate",
+    is_flag=True,
+    default=True,
+    help="(default enabled) Validate that the agent loads locally before deploying",
```
hessamb, line 478: nit: 
```suggestion
    help="(default enabled) Validate that the agent loads in a local venv before deploying",
```

fixieai/cli/agent/commands.py
```diff
@@ -321,44 +379,64 @@ def _ensure_agent_updated(
     default=True,
     help="(default enabled) Reload automatically.",
 )
+@click.option(
+    "--venv/--no-venv",
+    "use_venv",
+    is_flag=True,
+    default=True,
```
hessamb, line 386: I have a feeling `--venv` should be set to false by default for serve. Since the expectation is that the user is managing their own venv when developing their agent and serving it here.

fixieai/cli/agent/loader.py
```diff
@@ -42,14 +43,20 @@ def load_agent_from_path(
...
-    The FIXIE_AGENT_PATH environment variable should be set if agent.yaml is not in the current directory.
     The FIXIE_REFRESH_AGENT_ID environment variable can be set to trigger a refresh on startup.
     """
-    _, impl = load_agent_from_path(os.getenv("FIXIE_AGENT_PATH", "."))
+    _, impl = load_agent_from_path(".")
     return impl.app(os.getenv("FIXIE_REFRESH_AGENT_ID"))
+
+
+if __name__ == "__main__":
+    # Load the agent to make sure it can successfully be loaded.
```
hessamb, line 61: Please expand on this comment that this script is run under a venv to validate the agent

hessamb (approved):
Nicely done.

fixieai/cli/agent/commands.py
petersalas: It's definitely a little aggressive -- my thinking was to try to frontload missing dependency errors as early as possible in the process so that ideally they're iterating by adding things to `requirements.txt` rather than via `pip install` (and testing in a similar environment to what will be deployed). But we also don't want people to bounce off this if they can't figure out how to develop with their dependency either.

If `--no-venv` is the default do you think it should `pip install` into whatever env they're running in? Right now it only `pip install`s into the venv.

hessamb: I imagine a typical dev workflow is that people will start writing code locally, testing their deps into their current environment and test it via fixie serve. To me as a user, it feels like fixie serve should use my current environment to serve the agent, and the requirements.txt should be for the deploy environment. The venv check at fixie deploy time is really neat in that we do a final check of requirements.txt before deploying.

Obviously these are taste and hypothesis. So I cannot feel strongly one way or the other. Consider this comment as optional. I’m happy to go with the other way if your instincts are saying that’s going to live longer.

PR #92, opened 2023-03-20
[deployments] Upload as compressed tarball instead of individual files

petersalas:
This changes `fixie deploy` to upload files as a single gzip-compressed tarball instead of individual files.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -2,13 +2,14 @@ import contextlib
 import functools
 import io
 import os
-import pathlib
 import random
 import re
 import shlex
+import tarfile
+import tempfile
 import urllib.request
 from contextlib import contextmanager
-from typing import BinaryIO, Dict, List
+from typing import List, Optional
 
 import click
 import rich.console as rich_console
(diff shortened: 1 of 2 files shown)


hessamb (approved):

fixieai/cli/agent/commands.py
```diff
@@ -400,44 +416,27 @@ def deploy(ctx, path, metadata_only):
...
-                    )
-                    continue
-
-                full_path = os.path.join(root, filename)
-                upload_path = os.path.join(
-                    "agent", os.path.relpath(full_path, deploy_root)
+        with tempfile.TemporaryFile() as tarball_file:
+            with tarfile.open(fileobj=tarball_file, mode="w:gz") as tarball:
+                deploy_root = os.path.dirname(path)
+                tarball.add(
```
hessamb, line 422: does this do the right thing on windows?

juberti (approved):

fixieai/cli/agent/commands.py
petersalas: It does indeed! Good question :)

PR #97, opened 2023-03-23
Allow CodeShotAgents to specify model/temperature/maximum_tokens

petersalas:
This allows `CodeShotAgent`s to specify model/temperature/maximum_tokens.


fixieai/agents/code_shot.py
```diff
@@ -30,6 +30,13 @@
 _AGENT_ID_JWT_CLAIM = "aid"
 
 
+@pydantic_dataclasses.dataclass
+class ModelSettings:
```
hessamb, line 34: please add docstring for this class and the fields knowing that it will get exported as docs

fixieai/agents/code_shot.py
```diff
@@ -30,6 +30,13 @@
...
+    model: Optional[str] = None
```
hessamb, line 35: I think it's good to leave this field as string, as opposed to an SDK enum that gets updated. But I think we should leave a pointer in the docstring to a docs.fixie.ai page that has the list of models that are available

fixieai/agents/code_shot.py
hessamb, line 34: Model is a over-used terminology, I think `LlmSettings` might deliver the meaning quicker and to the point

fixieai/agents/code_shot.py
```diff
@@ -97,6 +105,9 @@ def __init__(
         corpora: Optional[List[corpora.DocumentCorpus]] = None,
         conversational: bool = False,
         oauth_params: Optional[oauth.OAuthParams] = None,
+        model: Optional[str] = None,
```
hessamb, line 108: I think this ctor should get a `LlmSetting` object directly

petersalas: That was the route I went with initially but worried it made setting these too verbose -- but since these are advanced settings anyway I think it's okay.

juberti (approved):

PR #100, opened 2023-03-27
Add response chunking and auth token verification to `GET /`

petersalas:
This adds response chunking and auth token verification to `GET /`. For auth token verification, in addition to verifying that the response has a valid Fixie request token we also verify that the agent ID in the token is the expected ID.

To avoid requiring that the agent ID be embedded in the _code_, the `fixie serve` and `fixie deploy` commands set the `FIXIE_ALLOWED_AGENT_ID` environment variable. The latter loads an `.env` file in the bootstrapping code. This environment variable is optional to avoid it being a breaking change -- we can make it required in the future. To avoid confusion between multiple uses of `agent_id` in `CodeShotAgent` this moves the refresh-on-startup logic entirely out of the agent implementation and into the agent loader where `fixie serve` can continue to use it.

Finally, to allow agents to be deployed against non-prod Fixie environments, this uses the new `.env` mechanism to set `FIXIE_API_URL` to match the environment that the agent is being deployed into.

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -4,14 +4,16 @@ import dataclasses
 import functools
 import inspect
 import json
+import logging
+import os
 import re
-import threading
-import time
+import warnings
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 import fastapi
+import fastapi.security
 import jwt
-import requests
+import starlette.responses
 import uvicorn
 import yaml
 from pydantic import dataclasses as pydantic_dataclasses
@@ -28,6 +30,10 @@ ACCEPTED_FUNC_NAMES = re.compile(r"^\w+$")
 
 # The JWT claim containing the agent ID
 _AGENT_ID_JWT_CLAIM = "aid"
+
+_RESPONSE_CHUNK_SIZE = 1024
+
+logger = logging.getLogger(__file__)
 
 
 @pydantic_dataclasses.dataclass
@@ -121,6 +127,7 @@ class CodeShotAgent:
         conversational: bool = False,
         oauth_params: Optional[oauth.OAuthParams] = None,
         llm_settings: Optional[LlmSettings] = None,
+        allowed_agent_id: Optional[str] = None,
     ):
         if isinstance(few_shots, str):
             few_shots = _split_few_shots(few_shots)
@@ -133,6 +140,12 @@ class CodeShotAgent:
         self.llm_settings = llm_settings
         self._funcs: Dict[str, Callable] = {}
         self._jwks_client = jwt.PyJWKClient(constants.FIXIE_JWKS_URL)
+        self._allowed_agent_id = allowed_agent_id or os.getenv("FIXIE_ALLOWED_AGENT_ID")
+        if self._allowed_agent_id is None:
+            warnings.warn(
+                "No allowed agent ID was specified, so your agent will accept requests intended for any agent. "
+                "Ensure that the FIXIE_ALLOWED_AGENT_ID variable is set to correct this."
+            )
 
         if oauth_params is not None:
             # Register default Funcs.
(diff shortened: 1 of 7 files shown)


hessamb (commented):
Nice PR!

Please see some comments inline, and bump the SDK version with these changes.

fixieai/agents/code_shot.py
```diff
@@ -307,7 +324,9 @@ class _VerifiedTokenClaims:
 
     @staticmethod
     def from_token(
-        token: str, jwks_client: jwt.PyJWKClient
+        token: str,
+        jwks_client: jwt.PyJWKClient,
+        allowed_agent_id: Optional[str],
```
juberti, line 329: maybe `expected_agent_id`

fixieai/agents/code_shot.py
```diff
@@ -121,6 +127,7 @@ def __init__(
         conversational: bool = False,
         oauth_params: Optional[oauth.OAuthParams] = None,
         llm_settings: Optional[LlmSettings] = None,
+        allowed_agent_id: Optional[str] = None,
```
juberti, line 130: The naming of this seems strange, should this just be `agent_id`?

juberti (commented):
Looks good overall, nice change. But I didn't totally follow this comment regarding agent_id in the PR description:

> To avoid confusion between multiple uses of agent_id in CodeShotAgent this moves the refresh-on-startup logic entirely out of the agent implementation and into the agent loader where fixie serve can continue to use it.

fixieai/cli/agent/commands.py
```diff
@@ -443,10 +453,15 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
 
 _DEPLOYMENT_BOOTSTRAP_SOURCE = """
 import os
-from fixieai.cli.agent import loader
+
+import dotenv
```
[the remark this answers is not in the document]
petersalas: Good question, it's now an explicit fixie-sdk dependency (turns out it was already in the transitive deps)

fixieai/agents/code_shot.py
```diff
@@ -232,7 +240,16 @@ def _handshake(self) -> fastapi.Response:
...
         )
         yaml_content = yaml.dump(dataclasses.asdict(metadata))
-        return fastapi.Response(yaml_content, media_type="application/yaml")
+
+        # Use a streaming response because prompts/fewshots can be very large.
+        chunks = (
+            yaml_content[start : start + _RESPONSE_CHUNK_SIZE]
+            for start in range(0, len(yaml_content), _RESPONSE_CHUNK_SIZE)
+        )
+        return starlette.responses.StreamingResponse(
```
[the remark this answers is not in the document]
petersalas: Nope, it just gets chunked in the response which satisfies GCR.

fixieai/agents/code_shot.py
petersalas: The intent was to try to make it explicitly narrow to both minimize dependencies and avoid users spending time setting this. (And also I didn't want to make it a breaking change.)

I really only added the argument for tests, but I could also just temporarily set the environment variable in the test to take it out of the API entirely?

petersalas:
> Looks good overall, nice change. But I didn't totally follow this comment regarding agent_id in the PR description:
> 
> > To avoid confusion between multiple uses of agent_id in CodeShotAgent this moves the refresh-on-startup logic entirely out of the agent implementation and into the agent loader where fixie serve can continue to use it.

@juberti basically there was already one optional `agent_id` scenario in `CodeShotAgent` for refresh-on-startup. Rather than have two slightly different but both optional `agent_id` arguments on different functions I moved the other one into the CLI side of things to avoid confusion. (The refresh on startup thing shouldn't be used/relied upon in the deployed code because it will happen ~randomly on scale-ups.)

fixieai/agents/code_shot.py
petersalas: I took it out of the API to try to remove it from the user's hands entirely -- lmk if you think that makes sense.

hessamb (approved):

juberti (approved):
Makes sense, thanks for explaining.

PR #102, opened 2023-03-27
Add support for storing multiple auth tokens in fixie auth

petersalas:
This adds support for multiple auth tokens being stored in the Fixie CLI config. The config contains a default API token as well as a number of URL-specific tokens:

- When looking for a token we first check if there's a token specific to the current Fixie URL and fall back to the default token
- When updating the token we add the token to the host-specific tokens list _as well as the default token_. This means that for any previously-unseen URL, the last token recorded by `fixie auth` will be used.

### DIFF
--- a/fixieai/cli/auth/commands.py
+++ b/fixieai/cli/auth/commands.py
@@ -13,9 +13,9 @@ def validate_not_authed(ctx, param, force):
         # --force is set
         return
     try:
-        api_key = user_config.load_config().auth_token
-        if api_key is not None:
-            username = _get_username(api_key)
+        auth_token = user_config.load_config().auth_token
+        if auth_token is not None:
+            username = _get_username(auth_token)
             click.echo("Already authenticated as ", nl=False)
             click.secho(username, fg="green", nl=False, bold=True)
             click.echo(". Set --force to force re-authentication.")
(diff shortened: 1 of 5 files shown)


mdwelsh (approved):
Nice!

fixieai/cli/auth/commands.py
```diff
@@ -12,14 +13,14 @@ def validate_not_authed(ctx, param, force):
         # --force is set
         return
     try:
-        config = user_config.load_config()
-        if config.fixie_api_key is not None:
-            username = _get_username(config.fixie_api_key)
+        api_key = user_config.load_config().api_key
```
juberti, line 16: FYI that these are auth_tokens, not api keys - let's name them accordingly here

fixieai/cli/auth/user_config.py
```diff
@@ -1,15 +1,51 @@
...
     fixie_api_key: Optional[str] = None
+    """The default API key, will be used if no URL-specific key matches."""
+
+    environment_api_keys: List[EnvironmentSpecificKey] = dataclasses.field(
+        default_factory=lambda: [],
+    )
+    """Environment-specific API keys."""
+
+    @property
+    def api_key(self) -> Optional[str]:
```
juberti, line 33: see above on naming

fixieai/constants.py
```diff
@@ -42,10 +40,11 @@ def fixie_api_key() -> str:
     if "FIXIE_API_KEY" in os.environ:
         return os.environ["FIXIE_API_KEY"]
     try:
-        with open(FIXIE_CONFIG_PATH) as fp:
-            api_key = yaml.safe_load(fp)["fixie_api_key"]
-            assert isinstance(api_key, str)
-            return api_key
+        from fixieai.cli.auth import user_config
+
+        api_key = user_config.load_config().api_key
```
juberti, line 45: see above on naming

juberti (approved):
LG but please adjust the naming to avoid confusion in the future

PR #103, opened 2023-03-29
Updated the OAuth example

Klaushayan:
To initialize a `@dataclass` you need to pass it values for its fields (that are not assigned a default value on class definition), otherwise, the generated `__init__` function throws a missing positional argument error.


juberti (approved):

juberti:
Thank you for the patch!

PR #104, opened 2023-03-29
Add standalone agent support

petersalas:
This adds a new agent type (`StandaloneAgent`) which can handle queries directly without any base prompt or fewshots.

The bulk of the change is factoring out most of `CodeShotAgent` into an `AgentBase` type that is now shared between `CodeShotAgent` and `StandaloneAgent`.

In addition to supporting registered funcs, `StandaloneAgents` can accept queries directly at `POST /`.


fixieai/agents/metadata.py
```diff
@@ -0,0 +1,29 @@
+from typing import List, Literal, Optional, Union
+
+from pydantic import dataclasses as pydantic_dataclasses
+
+from fixieai.agents import corpora as fixie_corpora
+from fixieai.agents import llm_settings
+
+
+@pydantic_dataclasses.dataclass
+class StandaloneAgentMetadata:
```
juberti, line 10: should these live within the respective agent files?

juberti (approved):
Nice PR! See my one minor comment.

fixieai/agents/metadata.py
petersalas: I put them here to avoid circular imports because the abstract method returns `Union[Standalone, CodeShot]`.

PR #109, opened 2023-03-31
fixed valid empty first line of fewshot bug

kolserdav:
When the first `fewshot` line is empty:
```python
FEW_SHOTS = """

Q: time
Ask Func[get_time]: input
Func[get_time] says: output
A: Now time is output
"""
```
, while run `fixie agent deploy` an incomprehensible error appeared:
```sh
...
Requirements installed successfully.
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/kol/Projects/fixie-agent/.fixie.venv/lib/python3.10/site-packages/fixieai/cli/agent/loader.py", line 95, in <module>
    load_agent_from_path(".")
  File "/home/kol/Projects/fixie-agent/.fixie.venv/lib/python3.10/site-packages/fixieai/cli/agent/loader.py", line 59, in load_agent_from_path
    agent_impl.validate()
  File "/home/kol/Projects/fixie-agent/.fixie.venv/lib/python3.10/site-packages/fixieai/agents/code_shot.py", line 90, in validate
    utils.validate_code_shot_agent(self)
  File "/home/kol/Projects/fixie-agent/.fixie.venv/lib/python3.10/site-packages/fixieai/agents/utils.py", line 34, in validate_code_shot_agent
    _validate_few_shot_prompt(
  File "/home/kol/Projects/fixie-agent/.fixie.venv/lib/python3.10/site-packages/fixieai/agents/utils.py", line 201, in _validate_few_shot_prompt
    pattern_matches[0] is not None
IndexError: list index out of range
Agent "trying" could not be loaded. If your agent requires additional dependencies make sure they are specified in ./requirements.txt.

```
Because this line https://github.com/fixie-ai/fixie-sdk/blob/877656079d31af882ec502e124a00592ab3c7b61/fixieai/agents/utils.py#L201 tried to access a non-existent object.

 This PR fixes this bug.

### DIFF
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -92,8 +92,14 @@ class CodeShotAgent(agent_base.AgentBase):
 
 def _split_few_shots(few_shots: str) -> List[str]:
     """Split a long string of all few-shots into a list of few-shot strings."""
-    lines = few_shots.splitlines()
+    # First, strip all lines to remove bad spaces.
+    few_shots = "\n".join(line.strip() for line in few_shots.splitlines())
+    # Then, split by "\n\nQ:".
+    few_shot_splits = few_shots.split("\n\nQ:")
+    few_shot_splits = [few_shot_splits[0]] + [
+        "Q:" + few_shot for few_shot in few_shot_splits[1:]
+    ]
 
     def not_empty(s):
         return s != ""
-    return list(filter(not_empty, lines))
+    return list(filter(not_empty, few_shot_splits))
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -100,6 +100,4 @@ def _split_few_shots(few_shots: str) -> List[str]:
         "Q:" + few_shot for few_shot in few_shot_splits[1:]
     ]
 
-    def not_empty(s):
-        return s != ""
-    return list(filter(not_empty, few_shot_splits))
+    return [few_shot for few_shot in few_shot_splits if few_shot]
--- a/fixieai/agents/code_shot.py
+++ b/fixieai/agents/code_shot.py
@@ -92,11 +92,8 @@ class CodeShotAgent(agent_base.AgentBase):
 
 def _split_few_shots(few_shots: str) -> List[str]:
     """Split a long string of all few-shots into a list of few-shot strings."""
-    # First, strip all lines to remove bad spaces.
-    few_shots = "\n".join(line.strip() for line in few_shots.splitlines())
-    # Then, split by "\n\nQ:".
-    few_shot_splits = few_shots.split("\n\nQ:")
-    few_shot_splits = [few_shot_splits[0]] + [
-        "Q:" + few_shot for few_shot in few_shot_splits[1:]
-    ]
-    return few_shot_splits
+    lines = few_shots.splitlines()
+
+    def not_empty(s):
+        return s != ""
+    return list(filter(not_empty, lines))
--- a/fixieai/agents/utils.py
+++ b/fixieai/agents/utils.py
@@ -22,6 +22,7 @@ EMBED_REF = re.compile(ODD_NUM_POUNDS.pattern + r"(?P<embed_key>\w+)(?!\w)")
 def strip_prompt_lines(agent: code_shot.CodeShotAgent):
     """Strips all prompt lines."""
     agent.base_prompt = _strip_all_lines(agent.base_prompt)
+
     for i, fewshot in enumerate(agent.few_shots):
         agent.few_shots[i] = _strip_all_lines(fewshot)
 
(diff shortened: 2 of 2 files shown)


juberti:
Thanks for the PR! Surprised we didn't automatically strip that blank line from the few-shot. Will take a closer look

hessamb:
We do make an attempt to strip fewshots from whitespaces, but apparently a bit too late. We first split the fewshots in [`_split_few_shots`](https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/code_shot.py#L70) and then strip each individual fewshot in [`utils.strip_prompt_lines`](https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/code_shot.py#L78). However, with the 2 extra newlines at the beginning of the fewshot we're forming an empty fewshot.

@kolserdav Thanks for the PR! I think the right fix would be to strip the fewshots at the beginning, before splitting them, here in [`_split_few_shots`](https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/code_shot.py#L93).

kolserdav:
1.  Fixed bug in method https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/code_shot.py#L93
2.There was confusion in method `_validate_base_prompt`, argument `prompt` was expected to be one line  https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/utils.py#L33 of `fewshot`, but in some places https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/utils.py#L275 it was treated as a full `fewshot`.

kolserdav:
I left all the checks that were before, only removed one duplicate extra check https://github.com/fixie-ai/fixie-sdk/blob/2dd09ddc5ecc789edbe96472eadc11d6dad0784b/fixieai/agents/utils.py#L291-302

fixieai/agents/code_shot.py
```diff
@@ -92,11 +92,8 @@ def validate(self):
...
-    few_shot_splits = few_shots.split("\n\nQ:")
-    few_shot_splits = [few_shot_splits[0]] + [
-        "Q:" + few_shot for few_shot in few_shot_splits[1:]
-    ]
-    return few_shot_splits
+    lines = few_shots.splitlines()
+
+    def not_empty(s):
+        return s != ""
+    return list(filter(not_empty, lines))
```
hessamb, line 99: We should not split to lines here. This function's job is to split a long list of fewshots into a list of individual fewshots. These are the 2 style of arguments CodeShotAgent accept (i.e. fewshots arguments of CodeShotAgent is `Union[str, List[str]]`).

The problem that we had before was that this split was making an empty "" fewshot. I think we only need to fix that problem.

kolserdav: I rolled back all the changes and only added a check for an empty string in `_split_few_shots`

fixieai/agents/utils.py
```diff
@@ -304,4 +304,4 @@ def _validate_few_shot_prompt(
 
 def _assert(condition: bool, msg: str, prompt: str):
     if not condition:
-        raise ValueError(f"{msg} in few-shot prompt: {prompt!r}.")
+        raise ValueError(f"{msg} in few-shot prompt: {prompt!r}.")
```
hessamb, line 307: Please revert changes to this file.

fixieai/agents/code_shot.py
```diff
@@ -99,4 +99,7 @@ def _split_few_shots(few_shots: str) -> List[str]:
     few_shot_splits = [few_shot_splits[0]] + [
         "Q:" + few_shot for few_shot in few_shot_splits[1:]
     ]
-    return few_shot_splits
+
+    def not_empty(s):
+        return s != ""
+    return list(filter(not_empty, few_shot_splits))
```
hessamb, line 105: The pythonic way to write this, would be:
```python
return [few_shot for few_shot in few_shot_splits if few_shot]
```
List comprehensions are as fast as filter and map, and they are more a bit more readable.

hessamb (commented):
Thanks @kolserdav. A couple of nits and then this PR is ready to go.

fixieai/agents/utils.py
kolserdav: Hmm, not sure why this is considered changed. In general, some unrelated changes could appear because I reverted the changes in the wrong way, - I copied the files from branch main, and then added the necessary changes to them, and I did not directly touch this line.

hessamb: Understandably this kind of change can get introduced. However, it's nice to revert the unintended changes. For example, we want to end files in newline, otherwise the lint checks (`just check`) fails. Or the newline that was removed from docstring above makes docstring harder to read. You could revert the uninteded changes via `get merge main && git checkout main -- /file/to/revert`.

PR #110, opened 2023-03-31
Update agent URL when tunnel URL changes

petersalas:
localhost.run tunnels periodically/randomly get recycled. This change updates the agent URL when it does.


petersalas:
Part/all of the problem seems to be attributed to using the `nokey` username: https://localhost.run/docs/faq#my-tunnel-name-changes-every-time-i-connect

I'm going to poke around with not using `nokey` and see if that helps before proceeding with this.

juberti (approved):
Seems like a smart approach.

PR #111, opened 2023-04-01
Add support for publishing/unpublishing agents.

mdwelsh:
Also adds support for getting current user info via `fixie user`, which is helpful to ensure that one is authenticated properly.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -133,18 +133,12 @@ def _update_agent_requirements(
     callback=_validate_url,
 )
 @click.option(
-    "--public",
-    prompt=True,
-    default=lambda: _current_config().public,
-    type=click.BOOL,
-)
-@click.option(
     "--requirement",
     multiple=True,
     type=str,
     help="Additional requirements for requirements.txt. Can be specified multiple times.",
 )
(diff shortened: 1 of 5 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -538,11 +532,22 @@ def _add_text_file_to_tarfile(path: str, text: str, tarball: tarfile.TarFile):
...
-def deploy(ctx, path, metadata_only, validate):
+def deploy(ctx, path, metadata_only, public, validate):
     console = rich_console.Console(soft_wrap=True)
     config = agent_config.load_config(path)
+    if config.public is not None:
+        console.print(
+            "[yellow]Warning:[/] The [blue]`public`[/] option in the agent config is deprecated and will be removed soon."
+        )
+        console.print(
+            "[yellow]Warning:[/] Use [blue]`fixie agent deploy --public`[/] instead."
```
zkoch, line 543: Maybe I'm confused, but is the correct syntax `fixie agent deploy --public` or `fixie agent publish`? Or both?

mdwelsh: Both. You can use `fixie agent deploy --public` to deploy+publish in one step, or use `fixie agent publish` on an already-deployed agent.

zkoch (approved):

PR #114, opened 2023-04-03
Add "fixie query" CLI command.

mdwelsh:
This is useful for doing a one-off query in an ephemeral session. I plan to use this in the GitHub Actions run for deploying agents from the fixie-examples repo, to send off a quick query to each deployed agent to make sure it is working correctly.

### DIFF
--- a/fixieai/cli/cli.py
+++ b/fixieai/cli/cli.py
@@ -51,6 +51,7 @@ fixie.add_command(agent_commands.serve, "serve")
 fixie.add_command(agent_commands.publish, "publish")
 fixie.add_command(agent_commands.unpublish, "unpublish")
 fixie.add_command(session_commands.new_session, "console")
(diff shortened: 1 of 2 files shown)


juberti (commented):
Nice addition, but maybe make this a --query parameter to fixie session new/open so we don't need to copy pasta as much?

mdwelsh:
> Nice addition, but maybe make this a --query parameter to fixie session new/open so we don't need to copy pasta as much?

That was my original thought, but it's not the same thing -- `fixie session new` is running the console, this is just sending a query and getting back the response (I don't want or need the console for this). So then you end up with a bunch of if statements based on whether you want the console or not - it's uglier.

mdwelsh:
I refactored this so there is a single base method used by the `session new`, `session open`, and `session query` commands. PTAL!

PR #115, opened 2023-04-04
Import from https://github.com/fixie-ai/fixie-sdk-ts

NickHeiner:
This is (mostly) a straight import from https://github.com/fixie-ai/fixie-sdk-ts. I didn't want to commingle any actual changes with the import.

Note that this PR is to `feature-ts`, not `main`.

I don't know that `fixieai/ts` is the ultimate directory for this to live in. We may evolve that as the structure comes into focus.


juberti (approved):
fixieai/agents-ts is what first occurred to me, but I agree we can tweak this later.

PR #117, opened 2023-04-04
Change formatting

NickHeiner:
I immediately decided I wanted this to be formatted differently. :smile:

I didn't realize there was a `onlyMultiline` option available on the trailing commas.


zkoch (approved):
yay, no commas! :D

PR #118, opened 2023-04-04
Add ability to `init` a TS agent

NickHeiner:
I anticipate tweaking the templated file contents once I implement actual agent functionality.

PR #119, opened 2023-04-04
Init command: do not rerun if an `agent.yml` already exists

NickHeiner:
@petersalas and I agreed this was more trouble than it was worth / overly clever.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -114,7 +114,9 @@ def config_already_exists():
     except FileNotFoundError:
         return False
 
+
 default_config = agent_config.AgentConfig()
+
 
 @agent.command("init", help="Creates an agent.yaml file.")
 @click.option(
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -114,9 +114,7 @@ def config_already_exists():
     except FileNotFoundError:
         return False
 
-
 default_config = agent_config.AgentConfig()
-
 
 @agent.command("init", help="Creates an agent.yaml file.")
 @click.option(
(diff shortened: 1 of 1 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -169,7 +187,7 @@ def init_agent(handle, description, entry_point, more_info_url, requirement, lan
     if is_typescript:
         main_file_extension = ".ts"
         # Before merging to main, update this to match the Python version.
-        agent_template_url = "https://raw.githubusercontent.com/fixie-ai/fixie-sdk/nth/template-ts-agent/fixieai/agents-ts/src/template.ts"
+        agent_template_url = "https://raw.githubusercontent.com/fixie-ai/fixie-sdk/feature-ts/fixieai/agents-ts/src/template.ts"
```
NickHeiner, line 190: Whoops – should have changed this before merging #118.

fixieai/cli/agent/commands.py
```diff
@@ -107,17 +107,28 @@ def _update_agent_requirements(
...
 
+def config_already_exists():
+    try:
+        agent_config.load_config()
+        return True
+    except FileNotFoundError:
+        return False
+
+
+default_config = agent_config.AgentConfig()
```
petersalas, line 118: nit: consider either wrapping this in a function (and doing `default_config().handle` etc.) or just inlining `agent_config.AgentConfig().handle` to avoid a slightly confusing (mutable) variable in the global scope

petersalas (approved):

fixieai/cli/agent/commands.py
```diff
@@ -150,6 +161,13 @@ def _update_agent_requirements(
     help="Additional requirements for requirements.txt. Can be specified multiple times.",
 )
 def init_agent(handle, description, entry_point, more_info_url, requirement, language):
+    if config_already_exists():
```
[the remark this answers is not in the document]
NickHeiner: Discussed in person: we think this is too sneaky to be worth it now, but we'll revisit if needed.

PR #124, opened 2023-04-05
Stream responses from standalone agents

juberti:
This allows standalone agents to stream their internal LLM state, similar to how our internal agents handle this in query_gen.


fixieai/cli/agent/commands.py
```diff
@@ -431,6 +431,8 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
 
         # Trigger an agent refresh each time it reloads.
         agent_env[FIXIE_REFRESH_ON_STARTUP] = "true"
+        agent_env["FIXIE_REFRESH_AGENT_ID"] = agent_api.agent_id
```
mdwelsh, line 434: What is this used for in this PR?

mdwelsh (approved):

fixieai/cli/agent/commands.py
juberti: looks like a bad merge, will revert

fixieai/agents/standalone.py
```diff
@@ -31,7 +33,7 @@ def __init__(
 
         self._handle_message = handle_message
         self._sample_queries = sample_queries
-        utils.validate_registered_pyfunc(handle_message, self)
+        utils.validate_registered_pyfunc(handle_message, self, is_generator=True)
```
petersalas, line 36: For my own understanding, is it a _requirement_ that the function be a generator/return an iterator? My only concern would be if this will trip folks up about needing to `yield` instead of `return` and so I'm wondering if there's a sensible way to tell if the function has returned a generator. (And/or infer it as such from the type signature)

juberti: It's not a requirement, but we'd need to handle both cases differently. (Which admittedly should be fairly straightforward from the typing).

juberti: Done.

fixieai/cli/agent/commands.py
```diff
@@ -443,9 +443,7 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
                 str(port),
                 "--factory",
             ]
-            + ["--reload"]
-            if reload
-            else [],
+            + (["--reload"] if reload else []),
```
petersalas, line 446: Weird that this is showing as a diff since this is already fixed in main

petersalas (approved):

fixieai/cli/agent/commands.py
juberti: synced with main and it's fixed now

PR #125, opened 2023-04-05
Add NodeJS server

NickHeiner:
Add a NodeJS server that can be executed both locally and deployed to Google Cloud Run.

# Demo
1. Run `poetry run fixie init --language ts`.
2. Run `poetry run fixie serve agent.yaml`. This command should output something like:
    ```
    Opening tunnel to 0.0.0.0:8181 via localhost.run.
    This replaces any existing deployment. Run fixie deploy to redeploy to prod.
    🦊 Agent nick1/test-agent-init-2 running locally on 0.0.0.0:8181, served via https://729c4028865cb3.lhr.life. 
    To interact with this agent, open another terminal window, and run: `fixie console --agent nick1/test-agent-init-2`.`
    ```
3. Run the `fixie console` command suggested in step (2).
4. See that you can interact with your agent.

# Out of scope / follow-up work
* watch mode
* better logging for the server
* support Agent functions being async
* @petersalas and I had a hard time getting the original code shots from the Python template to work. I think they were too vague. When we used the dice code shots, it worked immediately. So perhaps we should update the Python template to be dice as well.

# Design Notes
I briefly considered using `oclif` instead of `yargs`, but `oclif` has way more complexity than we need.


fixieai/agents-ts/.eslintrc.cjs
```diff
@@ -6,7 +8,7 @@ module.exports = {
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    project: 'tsconfig.json',
+    project: path.join(__dirname__, 'tsconfig.json'),
```
NickHeiner, line 11: This fixes an issue where VSCode was looking in the project root for the `tsconfig.json`, which isn't the right place.

fixieai/agents-ts/package.json
```diff
@@ -15,8 +15,11 @@
     "format": "dprint fmt",
     "format:check": "dprint check",
     "postinstall": "npm run build",
-    "build": "tsc -p .",
+    "build": "tsc -p . && chmod a+x dist/serve-bin.js",
     "start:dev": "npx nodemon",
+    "bin:dev": "npx ts-node src/serve-bin.ts",
+    "unit": "npx jest --runInBand",
```
NickHeiner, line 21: I had to do `runInBand` because otherwise the tests will try to launch servers on the same port.

fixieai/agents-ts/package.json
```diff
@@ -27,19 +30,34 @@
...
     "typescript": "^5.0.3"
   },
   "dependencies": {
     "@types/express": "^4.17.17",
-    "express": "^4.18.2"
+    "bunyan": "^1.8.15",
+    "bunyan-format": "^0.2.1",
+    "bunyan-middleware": "^1.0.2",
+    "express": "^4.18.2",
+    "got": "^11.8.6",
```
NickHeiner, line 56: got@12 uses ESM in a way that's really hard for us to use, so we can just use got@11 instead.

fixieai/agents-ts/src/serve-bin.ts
```diff
@@ -0,0 +1,82 @@
+#! /usr/bin/env node
+
+import fs from 'fs';
+import yaml from 'js-yaml';
+import _ from 'lodash';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import serve, { AgentConfig } from './serve';
+
+const { argv } = yargs(hideBin(process.argv))
```
NickHeiner, line 10: The error handling in this file will probably only be encountered by Fixie employees, not by third party devs. But I tried to make it reasonable anyway, to simplify debugging.

fixieai/agents-ts/src/serve.ts
```diff
@@ -0,0 +1,213 @@
...
+ */
+// @ts-expect-error
+if (!process[Symbol.for('ts-node.register.instance')]) {
+  tsNode.register();
+}
+
+/**
+ * This should be kept in sync with the `AgentConfig` dataclass.
+ */
+export interface AgentConfig {
```
NickHeiner, line 31: We don't actually need all these fields from TS, so arguably I could simplify this type.

fixieai/agents-ts/src/serve.ts
```diff
@@ -0,0 +1,213 @@
...
+  name?: string;
+  description: string;
+  more_info_url: string;
+  language: string;
+  entry_point: string;
+  deployment_url?: string;
+  public?: boolean;
+}
+
+interface AgentMetadata {
```
NickHeiner, line 42: It would be great if we could share all these types from somewhere, but :shrug:.

fixieai/agents-ts/src/serve.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  app.get('/', (_req, res) => res.send(agentRunner.getAgentMetadata()));
+  app.post('/:funcName', (req, res) => {
+    const funcName = req.params.funcName;
+    const body = req.body;
+
+    if (typeof req.body.message?.text !== 'string') {
+      res
+        .status(400)
+        // Is it a security problem to stringify untrusted input?
```
NickHeiner, line 176: I think the answer is "no" but I haven't fully thought about it. :smile:

fixieai/cli/agent/commands.py
```diff
@@ -167,14 +167,11 @@ def init_agent(handle, description, entry_point, more_info_url, requirement, lan
...
-    try:
-        current_config = agent_config.load_config()
-    except FileNotFoundError:
-        current_config = agent_config.AgentConfig()
+    current_config = agent_config.AgentConfig()
     current_config.handle = handle
     current_config.description = description
     current_config.entry_point = entry_point
-    current_config.language = language
+    current_config.language = "typescript" if is_typescript else "python"
```
NickHeiner, line 174: If we just set `current_config.language = language`, then we could end up with it being `ts` or `typescript` or `py` or `python`. This normalizes it.

fixieai/cli/agent/commands.py
```diff
@@ -167,14 +167,11 @@ def init_agent(handle, description, entry_point, more_info_url, requirement, lan
 
     is_typescript = language.lower() in ["typescript", "ts"]
 
-    try:
-        current_config = agent_config.load_config()
-    except FileNotFoundError:
-        current_config = agent_config.AgentConfig()
+    current_config = agent_config.AgentConfig()
```
NickHeiner, line 170: I realized we could simplify here, since the command no longer operates on a current config.

fixieai/cli/agent/commands.py
```diff
@@ -184,12 +181,12 @@ def init_agent(handle, description, entry_point, more_info_url, requirement, lan
     if is_typescript:
         main_file_extension = ".ts"
         # Before merging to main, update this to match the Python version.
-        agent_template_url = "https://raw.githubusercontent.com/fixie-ai/fixie-sdk/feature-ts/fixieai/agents-ts/src/template.ts"
+        agent_template_url = "https://raw.githubusercontent.com/NickHeiner/fixie-sdk/nth/nodejs-server/fixieai/agents-ts/src/template.ts"
```
NickHeiner, line 184: If I don't make this change, the demo won't work. :Smile:

fixieai/cli/agent/commands.py
```diff
@@ -451,18 +448,26 @@ def _validate_agent_loads_or_exit(
 )
 @click.option(
     "--venv/--no-venv",
-    "use_venv",
+    "use_venv_flag",
     is_flag=True,
     default=False,
-    help="Run from virtual environment",
+    help="Run from virtual environment [Only used for Python agents]",
```
NickHeiner, line 454: I think we'll continue to have awkwardness around flags that are only relevant for one language, but we can iron that out later.

fixieai/cli/agent/commands.py
```diff
@@ -474,16 +479,11 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
 
         agent_env[FIXIE_ALLOWED_AGENT_ID] = agent_id
 
-        # Validate that the agent loads before setting up the server.
-        _validate_agent_loads_or_exit(
-            console, agent_dir, config.handle, python_exe, agent_env
-        )
-
```
NickHeiner, line 481: Yanked this after discussing with @petersalas.

fixieai/cli/agent/commands.py
```diff
@@ -474,16 +479,11 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
...
         if use_tunnel:
             # Start up a tunnel via localhost.run.
             console.print(f"Opening tunnel to {host}:{port} via localhost.run.")
             console.print(
-                f"[yellow]This replaces any existing deployment, run [bold]fixie deploy[/bold] to redeploy to prod.[/yellow]"
+                f"[yellow]This replaces any existing deployment. Run [bold]fixie deploy[/bold] to redeploy to prod.[/yellow]"
```
NickHeiner, line 486: Fixing the grammar 😅

fixieai/cli/agent/commands.py
```diff
@@ -497,36 +497,52 @@ def _watch_tunnel():
...
+                        f"[yellow]Tunnel URL was updated. Now serving via {url}[/yellow]"
                     )
 
             threading.Thread(target=_watch_tunnel, daemon=True).start()
         else:
             agent_api = _ensure_agent_updated(client, agent_id, config)
 
         console.print(
-            f"🦊 Agent [green]{agent_api.agent_id}[/] running locally on {host}:{port}, served via {agent_api.func_url}"
+            f"🦊 Agent [green]{agent_api.agent_id}[/] running locally on {host}:{port}, served via {agent_api.func_url}. To interact with this agent, open another terminal window, and run: `fixie console --agent {agent_api.agent_id}`.`"
```
NickHeiner, line 508: I think this is the type of messaging we should have more of – we should hold the dev's hand through every step of the process.

fixieai/cli/agent/commands.py
```diff
@@ -497,36 +497,52 @@ def _watch_tunnel():
...
-            env=agent_env,
-            cwd=agent_dir or None,
-        ).check_returncode()
-
+                "--agent",
+                path,
+                "--silentStartup",
+                "--humanReadableLogs",
+                "--refreshMetadataAPIUrl",
+                client.get_refresh_agent_url(config.handle),
```
NickHeiner: It seems better to let the Python code own this value, rather than the Node re-deriving it.

fixieai/agents-ts/src/__tests__/serve.ts
```diff
@@ -0,0 +1,75 @@
+import fs from 'fs';
```
NickHeiner, line 1: Some of these tests were generated with help from https://www.codium.ai/.

mdwelsh (approved):
Nice work!

fixieai/cli/agent/commands.py
petersalas: We probably can make the `language` field an enum and/or just change its type to `Literal["typescript", "python"]` to let the type-checker verify.

fixieai/cli/agent/commands.py
petersalas: Good call. I think a pretty slick UX would be to make a TUI that splits the console and shows logs on the right and automatically starts a new session with the agent on the left. But it quickly spirals out of control with UX design problems :)

fixieai/agents-ts/src/template.ts
```diff
@@ -9,26 +9,27 @@
  *     https://github.com/fixie-ai/fixie-examples
  */
 
-// @ts-expect-error
-import { CodeShotAgent, Message } from 'fixieai';
+export const BASE_PROMPT = `I'm an agent that rolls virtual dice!`;
```
petersalas, line 12: We should probably align the Python/TS templates -- @hessamb might have ideas on how to make the template easier to reason about if it's run without changes.

fixieai/agents-ts/src/template.ts
```diff
@@ -9,26 +9,27 @@
...
 
-const agent = new CodeShotAgent(BASE_PROMPT, FEW_SHOTS);
-
-agent.registerFunc((query: Message): string => {
-  if (query.text === 'input') {
-    return 'output';
-  }
-  throw new Error('Invalid input');
-});
+export function roll(query: { text: string; }): string {
```
petersalas, line 31: I like the module-as-agent paradigm!

Since there's a handful of other config options as well, I'm wondering if we should wrap all the config in a single object, like `const export config = { base_prompt: "...", few_shots: "...", conversational: true }`, WDYT?

The other thing to consider is how users would configure a standalone agent, but I think mirroring the `GET /` response with a `"type"` field seems probably sufficient?

fixieai/agents-ts/package.json
```diff
@@ -15,8 +15,11 @@
     "format": "dprint fmt",
     "format:check": "dprint check",
     "postinstall": "npm run build",
-    "build": "tsc -p .",
+    "build": "tsc -p . && chmod a+x dist/serve-bin.js",
```
petersalas, line 18: Certainly not something you need to focus on right now but we should try to support Windows assuming it's straightforward to do so. This bit me at the hackathon where like 99% of the code worked on Windows but a couple of critical pieces didn't for dumb reasons :)

fixieai/agents-ts/src/serve.ts
petersalas: Oof, yeah that's a pain. Wondering if we should change both Python/TS sides to align to a convention like `contracts/foo.{ts,py}` to make them explicitly symmetrical?

fixieai/agents-ts/src/serve.ts
```diff
@@ -0,0 +1,213 @@
...
+  silentStartup,
+  refreshMetadataAPIUrl,
+  silentRequestHandling = false,
+  humanReadableLogs = false,
+}: {
+  agentConfigPath: string;
+  agentConfig: AgentConfig;
+  port: number;
+  silentStartup: boolean;
+  refreshMetadataAPIUrl: string;
```
petersalas, line 130: Assuming this code will eventually be used in the cloud as well, I think this should be optional (we'll want `fixie serve` to use it on changes but not `fixie deploy` on startup).

fixieai/agents-ts/src/serve-bin.ts
```diff
@@ -0,0 +1,82 @@
...
+  .epilog(
+    'This is an internal tool used by the Fixie SDK. Developers building on Fixie are not intended to use this tool; they should use the `fixie` CLI tool instead.',
+  );
+
+/**
+ * This is a little dance to make TS happy. We know that argv is not a Promise, based on how we wrote the yargs code,
+ * based but TS doesn't.
+ */
+type ExcludePromiseType<T> = Exclude<T, Promise<any>>;
+const staticArgv = argv as ExcludePromiseType<typeof argv>;
```
petersalas, line 70: For my own understanding, is there a reason not to do `Promise.resolve(argv).then(staticArgv => serve(...))`?

petersalas (approved):
Nice!

fixieai/cli/agent/commands.py
NickHeiner: I tried to make it an enum two days ago and ran into a series of problems that made me scrap the attempt. But maybe someone better at Python can do it better. :smile:

fixieai/cli/agent/commands.py
NickHeiner: For sure! Asking someone to open a new terminal window feels silly.

Alternatively, maybe we want to get rid of `fixie console` and just push people towards using the webapp? Are there particular advantages/use-cases that `fixie console` offers that we can't do on the web?

fixieai/agents-ts/src/serve-bin.ts
NickHeiner: You're right that we could treat `argv` as a promise and get around this issue. But given that it's not a promise, it feels a little obfuscatory to do so. But it's a minor point and your approach would be pretty similar overall I think.

Also, we wouldn't want to use `then` – we'd want to use `await`.

fixieai/agents-ts/src/template.ts
NickHeiner: Discussed in person + [Slack](https://fixie-ai.slack.com/archives/C03H4FX54JZ/p1680803400136669). We'll defer making a change on this.

PR #127, opened 2023-04-06
Allow JS agents to provide async funcs

NickHeiner:
This is a table-stakes thing that JS devs will expect.

### DIFF
--- a/fixieai/agents-ts/src/__tests__/serve.ts
+++ b/fixieai/agents-ts/src/__tests__/serve.ts
@@ -68,19 +68,57 @@ describe('server starts', () => {
 
     expect(response.statusCode).toBe(404);
     expect(response.body).toBe(
-      'Function not found: function-does-not-exist. Functions available: roll, willThrowError',
+      'Function not found: function-does-not-exist. Functions available: roll, willThrowError, willThrowErrorAsync, rollAsync',
     );
   });
 
-  it('Function being called throws an error', async () => {
-    const response = await got.post(`http://localhost:${port}/willThrowError`, {
-      json: { message: { text: 'input' } },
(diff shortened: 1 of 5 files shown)


fixieai/agents-ts/src/__tests__/serve.ts
```diff
@@ -68,19 +68,57 @@ describe('server starts', () => {
 
     expect(response.statusCode).toBe(404);
     expect(response.body).toBe(
-      'Function not found: function-does-not-exist. Functions available: roll, willThrowError',
+      'Function not found: function-does-not-exist. Functions available: roll, willThrowError, willThrowErrorAsync, rollAsync',
     );
   });
 
-  it('Function being called throws an error', async () => {
```
NickHeiner, line 75: This diff looks bigger than it is – I'm just adding two new tests. Two previously-existing tests are moved into this `describe` block.

fixieai/agents-ts/src/serve.ts
```diff
@@ -165,41 +166,44 @@ export default async function serve({
...
-      logger.error(
-        { error: errorForLogging, functionName: funcName },
-        'Error running agent function',
-      );
-      res.status(500).send(errorForLogging);
-    }
-  });
+
+      try {
+        const result = await agentRunner.runFunction(funcName, body.message);
```
petersalas, line 190: TIL you can `await` non-promises too :)

petersalas (approved):

fixieai/agents-ts/src/serve.ts
NickHeiner: Yep! It normalizes to a promise.

PR #129, opened 2023-04-07
NodeJS deploy

NickHeiner:
This works alongside https://github.com/fixie-ai/fixie/pull/1048.


fixieai/agents-ts/package.json
```diff
@@ -1,20 +1,22 @@
 {
-  "name": "@fixie/sdk",
-  "private": true,
-  "version": "1.0.0",
+  "name": "@fixieai/sdk",
+  "version": "1.0.0-6",
```
NickHeiner, line 3: Unfortunately, I don't see a better way to dev/test these PRs without publishing a bunch of prerelease versions. Any way we work around this would be more noise than it's worth. The prerelease versions are a little annoying but I think almost no one will even notice.

fixieai/agents-ts/package.json
```diff
@@ -1,20 +1,22 @@
...
+  "bin": {
+    "serve-bin": "dist/serve-bin.js"
+  },
   "scripts": {
     "lint": "eslint . --max-warnings 0",
     "lint:fix": "eslint . --fix",
     "format": "dprint fmt",
     "format:check": "dprint check",
-    "postinstall": "npm run build",
+    "prepublishOnly": "npm run build",
```
NickHeiner, line 19: This was a mistake – by putting it as `postinstall`, I had consumers running `build`, which is not what we want.

fixieai/agents-ts/src/serve-bin.ts
```diff
@@ -35,6 +35,11 @@ const { argv } = yargs(hideBin(process.argv))
       required: true,
       default: 'config.yaml',
       coerce: (path: string): { parsed: AgentConfig; path: string; } => {
+        if (path === '.') {
+          // This is a hack to make this work with what's baked into the service.
+          // eslint-disable-next-line no-param-reassign
+          path = 'agent.yaml';
+        }
```
NickHeiner, line 42: Workaround for https://github.com/fixie-ai/fixie/pull/1048/files#r1160353921.

fixieai/agents-ts/src/serve.ts
```diff
@@ -23,6 +23,9 @@ import * as tsNode from 'ts-node';
 // @ts-expect-error
 // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
 if (!process[Symbol.for('ts-node.register.instance')]) {
+  /**
+   * We may need to explicitly pass the tsconfig.json here. Let's try omitting it and see if that works.
```
NickHeiner, line 27: It works for now! Maybe down the line we'll need to reevaluate.

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +644,74 @@ def deploy(ctx, path, metadata_only, public, validate):
...
     agent_dir = os.path.dirname(path) or "."
     if validate:
-        # Validate that the agent loads in a virtual environment before deploying.
-        python_exe, agent_env = _configure_venv(console, agent_dir)
-        agent_env[FIXIE_ALLOWED_AGENT_ID] = agent_id
-        _validate_agent_loads_or_exit(
-            console, agent_dir, config.handle, python_exe, agent_env
-        )
+        # Skip this step for TS. The only thing TS is validating is pieces that ultimately should end up in the agent.yml
+        if not is_typescript:
```
NickHeiner, line 648: I decided to skip this for TS entirely, because the only validation we'd do is of fields that are soon to be moved into yaml, hopefully. (And this provides another benefit to moving it into yaml – no need to implement the same validation in every client language.)

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +644,74 @@ def deploy(ctx, path, metadata_only, public, validate):
...
-                        }.items()
-                    ),
-                    tarball,
-                )
+            result.check_returncode()
+
+            tarball_name = result.stdout.splitlines()[-1]
+            path_of_tarball_to_upload = os.path.join(temp_dir, tarball_name)
+            tarball_file = open(path_of_tarball_to_upload, "rb")
+        else:
```
NickHeiner, line 681: This branch is unchanged – just moving it into the `else`.

fixieai/constants.py
```diff
@@ -4,7 +4,8 @@
 import os
 
 # Base Fixie platform URL.
-FIXIE_API_URL = os.getenv("FIXIE_API_URL", "https://app.fixie.ai")
+# FIXIE_API_URL = os.getenv("FIXIE_API_URL", "https://app.fixie.ai")
+FIXIE_API_URL = os.getenv("FIXIE_API_URL", "https://psalas-node-deployments-pgaenaxiea-uc.a.run.app")
```
NickHeiner, line 8: We'll remove this before merging.

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +644,74 @@ def deploy(ctx, path, metadata_only, public, validate):
...
+                            f"{key}={value}"
+                            for key, value in {
+                                FIXIE_ALLOWED_AGENT_ID: agent_id,
+                                "FIXIE_API_URL": constants.FIXIE_API_URL,
+                            }.items()
+                        ),
+                        tarball,
+                    )
+
+        tarball_file.seek(0)
```
NickHeiner, line 709: @petersalas when I run `just install`, I get this validation error:

```
fixieai/cli/agent/commands.py:683: error: Incompatible types in assignment (expression has type "IO[bytes]", variable has type "BufferedReader")  [assignment]
```

What's the best way to resolve this? In TS, we'd do something like:

```ts
let tarballFile: IO<Bytes> | BufferedReader
```

But I'm not sure in Python where to import those names from.

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +645,80 @@ def deploy(ctx, path, metadata_only, public, validate):
...
+            result = subprocess.run(
+                # If the dev adds a `files` field to their package.json, and that field excludes `agent.yaml`, then
+                # this the tarball will omit that file, which would break the deploy.
+                # We could add it to the tarball manually, or check the built tarball to see if it's there, etc.
+                # But for now, I think we can skip worrying about it.
+                [
+                    "npm",
+                    "pack",
+                    "--pack-destination",
+                    temp_dir,
```
petersalas, line 673: IIUC this is using the current directory -- we should probably align `serve` and `deploy` on how the CLI goes from `agent.yaml` -> `package.json`. Options that come to mind:
1. Require that it sit immediately next to `agent.yaml`
2. Allow it to be next to `agent.yaml` or in any ancestor directory (`findup` a la `fixie serve`)
3. Have a way for `agent.yaml` to point to the package relative to itself that the CLI code uses for both serve/deploy (`entry_point`?)

I don't feel super strongly about which one to use as long as `fixie serve` and `fixie deploy` are consistent. :)

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +645,80 @@ def deploy(ctx, path, metadata_only, public, validate):
...
-                        for key, value in {
-                            FIXIE_ALLOWED_AGENT_ID: agent_id,
-                            "FIXIE_API_URL": constants.FIXIE_API_URL,
-                        }.items()
-                    ),
-                    tarball,
-                )
+            result.check_returncode()
+
+            tarball_name = result.stdout.splitlines()[-1]
```
petersalas, line 681: Consider using `npm pack --json` and pulling the filename from `json.loads(result.stdout)[0]["filename"]`

petersalas (approved):

docs/ts.md
```diff
@@ -0,0 +1,47 @@
...
+
+* If your code is in TypeScript, it'll be loaded with [`ts-node`](https://typestrong.org/ts-node/). `ts-node` will look for a `tsconfig.json` in your project. The `@fixieai/sdk` dependency declared in your `package.json` will transitively pull in `ts-node`.
+    * If you want to control the TypeScript compilation yourself:
+        1. Create a `tsc` command that will emit JS.
+        1. Call that command from [the `prepublishOnly` hook](https://docs.npmjs.com/cli/v9/using-npm/scripts#prepare-and-prepublish).
+        1. Update your `package.json` `main` field to point to the emitted JS entrypoint.
+
+### Limitations
+
+* You can't pass [flags/env vars to NodeJS](https://nodejs.org/api/cli.html).
```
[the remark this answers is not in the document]
NickHeiner: Yes, I think that's a good idea. And maybe that'll be an easy way to pass both env vars to NodeJS and the funcs.

fixieai/cli/agent/commands.py
```diff
@@ -643,49 +645,80 @@ def deploy(ctx, path, metadata_only, public, validate):
...
-                    arcname="agent",
-                    filter=functools.partial(
-                        _tarinfo_filter, console, agent_dir, "agent/"
-                    ),
-                )
+        if is_typescript:
+            temp_dir = tempfile.gettempdir()
+            result = subprocess.run(
+                # If the dev adds a `files` field to their package.json, and that field excludes `agent.yaml`, then
+                # this the tarball will omit that file, which would break the deploy.
```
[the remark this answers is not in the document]
NickHeiner: Good point! Another benefit of your suggestion earlier to remove `serve-bin` from having knowledge of the yaml.

fixieai/cli/agent/commands.py
NickHeiner: Good callout – I'll just change this to be `findup` like what `serve` does.

fixieai/cli/agent/commands.py
NickHeiner: Great idea!

PR #137, opened 2023-04-10
Support embeds in NodeJS funcs

NickHeiner:
This handles both inbound embeds (e.g. a func receives a message with an embed) and outbound embeds (a func produces a message with embeds).

# Demo
image
[Session link](https://app.fixie.ai/sessions/clumsy-humdrum-puzzle)

This is the code for `Func[chart]`:

```ts
export async function chart(): Promise<Message> {
  const url = `https://quickchart.io/chart?c=${encodeURIComponent(JSON.stringify({
    type: "line",
    data: {
      labels: [
        "January",
        "February", 
        "March", 
        "April",
        "May",
        "June",
        "July",
        "August",
        "September",
        "October",
        "November",
        "December",
      ],
      datasets: [
        {
          label: "Average Temperature (°F)",
          data: [45, 47, 50, 54, 60, 65, 68, 67, 63, 56, 48, 45],
        },
      ],
    },
  }))
}`;
  return {
    text: 'here is your chart #chart',
    embeds: {
      chart: await Embed.fromUri('image/webp', url)
    }
  };
}
```


fixieai/agents-ts/src/__tests__/serve.ts
```diff
@@ -6,7 +6,7 @@ import tempy from 'tempy';
 import type { PromiseType } from 'utility-types';
 import serve from '../serve';
 
-const agentPackagePath = path.resolve(__dirname, '..', 'fixtures');
+const agentPackagePath = path.resolve(__dirname, '..', 'fixtures', 'normal');
```
NickHeiner, line 9: I split the test fixtures into `normal` and `watch`, because the `watch` mode tests need to be more constrained, but I didn't want those constraints to apply to the rest of the test suite.

Some of this diff is just noise related to that, but isn't actually a meaningful change.

fixieai/agents-ts/src/embed.ts
```diff
@@ -0,0 +1,73 @@
...
+    public readonly uri: string,
+  ) {
+  }
+
+  serialize(): SerializedEmbed {
+    return {
+      // This is to match the Python agent's API.
+      // eslint-disable-next-line camelcase
+      content_type: this.contentType,
+      uri: uriFromBase64(this.base64Data),
```
juberti, line 57: This doesn't have to be base64, it could be a typical URL.

fixieai/agents-ts/src/fixtures/normal/chart-data.txt
```diff
@@ -0,0 +1 @@
...0E9wHOccba9WjE+Y+R4AD+ZyuVd+/PHH+UOGDDmEmW+Aqjuf5HneTAAYP378OrFY7DgAexNRnJnnMPMNnuc91l3XYihdhBBHALgRwJeZTGaLsAhhQbstAZwL4FD8FNf8CGBzs/gD2Lb9JBHtBeUZv0NnzglCiNUAfApgjXw+v2ljY+Mn3TlPw5KYAN1gMBgMBoOhH+E4zkEAjmTmIcz8FRE9zcxPdiYcJ4S4GMBJzHyb53lV3TtbQz+FhBBPAdgTwFVSymWq/gshfg0lFve97/snpNPpz7pjkr2durq6bX3ffx1AFMBkKeUVxdo5jiOZ2QYwTUq5QsJyhq7DBOgGg8FgMBgMhk6xbXssEd1mAnRDd2Lb9jZE9AYAIqIRruv+u6fn1BdxHOcKbWv4ne/7W4TLTxoaGhKZTOZyZq6Hsk/cq7NddkP3YUTiDAaDwWAwGAzLYhQAWJb1eE9PxNB/8DzvbWaeBSDCzFfDbCyuFMx8NoAvAKxpWdbFwXHbtke3t7e/ooPzW3K53O9McN47MG90g8FgMBgMBkNRHMfZh5kfAvBRIpEYPn369PaenpOh/9DQ0LBGe3v7ewCGEtERruvevNxOhqWwbbuaiK6FEmuvAjCWiP4M4CtmPtnzvKYenqIhhAnQDQaDwWAwGPo3BGV7tsQxx3GOYeZLoFJj90qn0+/2wNwM/RzHcWqYOQ3g80QiscX06dO/7+k59UHIcZwXmHlX/dqHEuGbIqX8ugfnZShCtKcnYDAYDAaDwWDoGRzHEcxcR0STKioq/vXpp5+uEY1G92Pmk5l5ByJ6NZ/PV6XT6f/29FwN/ZOKiopr582bJ4jol+3t7VsB+GdPz6kPwsw8CcDLRPRCPp8/Jp1Ov9nTkzIUx+ygGwwGg8FgMPRDHMfZmplfBxDXh9oBJPTPLQAuBXC1qUs19DSO42yWy+W+aWxsnN/Tc+nLCCF2kVK+jKUzZgy9CBOgGwwGg8FgMPRTbNv+FRGdDGAXABYzv01EdyYSidmm3txgMBgMBoPBYDAYDAaDwWAwGAwGg8FgMBgMBoPBYDAYDAaDwWAwGAwGg8FgMBgMBoPBYDAYDAaDwWAwGAwGg8FgMBgMBoPBYDAYDAaDwWAwGAwGg8FgMBgMBoPBYDAYDAaDwWAwGAwGg8FgMBgMBoOh5/l/OY4pM3f/wK8AAAAASUVORK5CYII=
```
juberti, line 1: did you try converting this to png? Might be a lot smaller

fixieai/agents-ts/src/embed.ts
```diff
@@ -0,0 +1,97 @@
...
+     * If this embed was constructed from base64 data, then this URI will be a data URI. This URL is not guaranteed to point to an external resource.
+     */
+    public readonly uri: string,
+  ) {
+  }
+
+  /**
+   * Format this Embed for sending to the Fixie API.
+   */
+  serialize(): SerializedEmbed {
```
NickHeiner, line 52: @petersalas are there docs I can link to specifying the wire format?

fixieai/agents-ts/src/serve.ts
```diff
@@ -157,7 +163,13 @@ export default async function serve({
      * outside its directory, the watcher won't watch them. This is a potential rough edge but it's also the way
      * Nodemon works, and I think it's unlikely to be a problem in practice.
      */
-    watcher = chokidar.watch(absolutePackagePath).on('all', async () => {
+    watcher = chokidar.watch(absolutePackagePath, {
+      /**
+       * We may eventually want to change this to ignore all gitignores.
+       */
+      ignored: /node_modules/,
```
NickHeiner, line 170: Failing to ignore `node_modules` was causing OOMs.

fixieai/agents-ts/src/embed.ts
juberti: IOW you should be able to do either do fromBase64("image/png", <base64 data>) or Embed("image/png", "http://image.png") (and ideally with the last one the content type would be optional)

fixieai/agents-ts/src/serve.ts
```diff
@@ -157,7 +163,13 @@ export default async function serve({
...
+      ignoreInitial: true,
```
NickHeiner, line 171: This solves the "startup is spammy" issue I'd described below.

fixieai/agents-ts/src/serve.ts
```diff
@@ -216,32 +224,28 @@ export default async function serve({
...
-            res.status(404).send(e.message);
-            return;
-          }
-          const errorForLogging = _.pick(e, 'message', 'stack');
-          logger.error(
-            { error: errorForLogging, functionName: funcName },
-            'Error running agent function',
-          );
-          res.status(500).send(errorForLogging);
-        }
```
NickHeiner, line 244: This was a merge error from before. (But all the unit tests still passed, so I guess it didn't impact correctness. 🤪 )

This section should just have been removed.

fixieai/agents-ts/src/fixtures/normal/chart-data.txt
NickHeiner: I didn't try to optimize for file size at all. Do you think it matters?

fixieai/agents-ts/src/embed.ts
NickHeiner: Let's chat in person – my understanding from @petersalas was that the func should always emit an embed with a `base64` data URL.

NickHeiner: Pushed changes after chatting in person.

fixieai/agents-ts/src/embed.ts
```diff
@@ -0,0 +1,73 @@
...
+    return {
+      // This is to match the Python agent's API.
+      // eslint-disable-next-line camelcase
+      content_type: this.contentType,
+      uri: uriFromBase64(this.base64Data),
+    };
+  }
+
+  static fromBase64(contentType: string, base64Data: string): Embed {
```
juberti, line 84: I think you want fromText and fromBinary instead

fixieai/agents-ts/src/embed.ts
petersalas: I think we _could_ convert all URIs to data URIs in the agent but it's not necessary to do so. One catch that comes to mind is that it requires the contents to be loaded in memory and might cause OOMs. (Agent OOMs are better than Fixie OOMs, but the latter we can address by being deliberate about how we manage those requests.) So if we want to avoid having a hard limit on embed size we're probably better off passing them indirectly back to the service.

fixieai/agents-ts/src/serve.ts
```diff
@@ -214,34 +221,40 @@ export default async function serve({
...
+        const result: Partial<SerializedMessage> = _.pick(message, 'text');
+        result.embeds = _.mapValues(message.embeds, (e) => e.serialize());
+        return result as SerializedMessage;
+      }
+
+      async function messageOfSerializedMessage(serializedMessage: SerializedMessage): Promise<Message> {
+        const result: Partial<Message> = _.pick(serializedMessage, 'text');
+        const embedPromises = Object.entries(serializedMessage.embeds ?? {}).map(async (
+          [embedName, embed],
+        ) => [embedName, await Embed.fromUri(embed.content_type, embed.uri)]);
```
petersalas, line 237: IMO we should let the agent code initiate this if they want the data directly. (But they may not -- for example they may want to pass that URI down to another service or w/e)

fixieai/agents-ts/src/embed.ts
```diff
@@ -0,0 +1,111 @@
...
+    return response.body;
+  }
+
+  /**
+   * Create an embed from base64 data.
+   *
+   * @param contentType The MIME content type of the object, e.g., "image/png" or "application/json" (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+   * @param textData The base64-encoded data for this embed.
+   */
+  static fromBase64(contentType: string, textData: string): Embed {
```
petersalas, line 94: Seems like this should be `fromText` or `fromString` (Lines 102-103 seem to do the base64 conversion)

fixieai/agents-ts/src/embed.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    const base64 = Buffer.from(textData, 'utf-8').toString('base64');
+    const uri = uriFromBase64(base64);
+
+    return new Embed(contentType, uri);
+  }
+
+  static fromBinary(contentType: string, binaryData: Buffer): Embed {
+    return Embed.fromBase64(contentType, binaryData.toString('base64'));
```
petersalas: I think these would be more natural layered in the other direction, e.g.

`fromText = (contentType, text) => fromBinary(contentType, Buffer.from(text, 'utf-8'))`

PR #150, opened 2023-04-12
Add self-check for version update.

mdwelsh:
This PR adds code to show a warning to the user if there is a newer version of the `fixieai` package on PyPi when they run the CLI.

### DIFF
--- a/fixieai/cli/cli.py
+++ b/fixieai/cli/cli.py
@@ -3,6 +3,9 @@
 from typing import Optional
 
 import click
+import pkg_resources
+import requests
+from packaging import version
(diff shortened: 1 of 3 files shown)


fixieai/cli/cli.py
```diff
@@ -54,5 +57,40 @@ def fixie(ctx):
 fixie.add_command(session_commands.query, "query")
 
 
+def get_installed_version(package_name):
```
juberti, line 60: type hints? (here and below)

juberti (approved):
LG with nits

PR #151, opened 2023-04-12
Add OpenAI proxy for agents

petersalas:
This allows agents to use Fixie as an OpenAI proxy in the context of agent requests.

When an agent (that does not set either `OPENAI_API_{BASE,KEY}` in its `.env` file) is loaded, we:
1. Set `OPENAI_API_BASE` to point to the Fixie OpenAI proxy
2. Set `OPENAI_API_KEY` to a fake key value
3. Patch OpenAI's `ApiRequestor` object to replace the fake key with the auth token from the current request (for requests that would be sent to Fixie).

During request handling, middleware puts the auth token in a `ContextVar` that the patched `ApiRequestor` will read.

### DIFF
--- a/fixieai/agents/openai_proxy.py
+++ b/fixieai/agents/openai_proxy.py
@@ -1,12 +1,11 @@
 import contextvars
+import functools
 import logging
 import os
 import sys
 
-from starlette.middleware.base import BaseHTTPMiddleware
-
-import functools
 import wrapt
+from starlette.middleware.base import BaseHTTPMiddleware
 
 import fixieai.constants
 
--- a/fixieai/agents/openai_proxy.py
+++ b/fixieai/agents/openai_proxy.py
@@ -25,7 +25,7 @@ _current_request_token: contextvars.ContextVar[Optional[str]] = contextvars.Cont
 def enable_openai_proxy() -> bool:
     """Enables the Fixie OpenAI proxy for outbound OpenAI requests made in the context of an agent request.
 
-    Returns a value that indicates whether or not proxy is enabled. If either OPENAI_API_BASE or
+    Returns a value that indicates whether or not the proxy is enabled. If either OPENAI_API_BASE or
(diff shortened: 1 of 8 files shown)


fixieai/agents/openai_proxy.py
```diff
@@ -0,0 +1,65 @@
...
+        module.APIRequestor.__init__ = patched_init
+
+    wrapt.register_post_import_hook(_patch_api_requestor, "openai.api_requestor")
+    os.environ["OPENAI_API_BASE"] = fixieai.constants.FIXIE_OPENAI_PROXY_URL
+    os.environ["OPENAI_API_KEY"] = FAKE_OPENAI_KEY
+
+    return True
+
+
+class AuthTokenForwarderMiddleware(BaseHTTPMiddleware):
```
juberti, line 53: Docstring would be helpful here. I assume this sniffs out the auth token from the latest inbound call?

juberti (commented):
I like the approach, swapping the URL is clearly the cleanest solution.

juberti:
Is this ready for review?

fixieai/agents/openai_proxy.py
```diff
@@ -0,0 +1,109 @@
...
+_current_request_token: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
+    "openai_proxy.current_request_token",
+    default=None,
+)
+
+
+def enable_openai_proxy() -> bool:
+    """Enables the Fixie OpenAI proxy for outbound OpenAI requests made in the context of an agent request.
+
+    Returns a value that indicates whether or not proxy is enabled. If either OPENAI_API_BASE or
```
juberti, line 28:
```suggestion
    Returns a value that indicates whether or not the proxy is enabled. If either OPENAI_API_BASE or
```

juberti (approved):

PR #153, opened 2023-04-14
Make func signatures more permissive, allow entry_point to point directly to functions

petersalas:
This change makes func signatures more permissive to make it easier to use from codebases without a dependency on the Fixie SDK, namely:
1. `str` parameters are allowed
2. Parameter semantics are inferred by type first (then name)
3. A single untyped argument is assumed to be a `str` and will take the query

Additionally, the `agent.yaml` `entry_point` field can now point to _either_ an Agent implementation or a function. If the latter, it is wrapped in an `AgentFunc`/`StandaloneAgent`.

It also consolidates all func type logic (validation, argument coercion, return type coercion) into a new type (`AgentFunc`) that is independent of `AgentBase`. To make this a non-breaking change, untyped agent-registered funcs continue to default to `api.Message` while the agent-less funcs default to `str`.

### DIFF
--- a/fixieai/agents/agent_func.py
+++ b/fixieai/agents/agent_func.py
@@ -3,18 +3,43 @@ from __future__ import annotations
 import collections.abc
 import inspect
 import typing
-from typing import Any, Callable, Iterable, Optional, Tuple, Union, get_type_hints
+from typing import (
+    Any,
+    Callable,
+    Iterable,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    Union,
+    get_type_hints,
+)
 
 from fixieai.agents import api
 from fixieai.agents import oauth
 from fixieai.agents import user_storage
 
-# An ArgumentMapper converts the a query+agent ID into a supported func argument (e.g. "query" or "user_storage")
+# An ArgumentMapper converts a query+agent ID into a supported func argument (e.g. "query" or "user_storage")
 ArgumentMapper = Callable[[api.AgentQuery, str], Any]
+
+# BoundArgumentMapper is an ArgumentMapper that has been bound to a named parameter.
 BoundArgumentMapper = Tuple[str, ArgumentMapper]
 
 
 class AgentFunc:
+    """A Python function that can be invoked by Fixie.
+
+    Wrapped functions can take up to three arguments:
+
+    1. A message or query of type: `str`, `api.AgentMessage`, or `api.AgentQuery`;
+       name: "query", or "message"; or the first parameter if no other rules apply.
+    2. A user storage object of type `user_storage.UserStorage` or name "user_storage".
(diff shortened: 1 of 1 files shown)


fixieai/agents/agent_func.py
```diff
@@ -0,0 +1,226 @@
...
+from fixieai.agents import api
+from fixieai.agents import oauth
+from fixieai.agents import user_storage
+
+# An ArgumentMapper converts the a query+agent ID into a supported func argument (e.g. "query" or "user_storage")
+ArgumentMapper = Callable[[api.AgentQuery, str], Any]
+BoundArgumentMapper = Tuple[str, ArgumentMapper]
+
+
+class AgentFunc:
```
hessamb, line 17: Please add some docstring for this class and the `create` method

fixieai/agents/agent_func.py
```diff
@@ -0,0 +1,226 @@
...
+        if any(
+            param.kind in (param.VAR_KEYWORD, param.VAR_POSITIONAL)
+            for param in params.values()
+        ):
+            raise TypeError(
+                f"Registered function {func_name} cannot accept variable args: {params!r}."
+            )
+
+        # Ensure every parameter is used exactly once.
+        unresolved_parameters = set(name for name in params.keys())
```
hessamb, line 64: nit
```suggestion
        unresolved_parameters = set(params.keys())
```

fixieai/agents/agent_func.py
```diff
@@ -0,0 +1,226 @@
...
+
+        result = self._impl(**kwargs)
+        return _adapt_func_result(result, self._allow_multiple_responses)
+
+    @classmethod
+    def create(
+        cls,
+        func: Callable,
+        oauth_params: Optional[oauth.OAuthParams],
+        default_message_type: type,
```
hessamb, line 43:
```suggestion
        default_message_type: Union[Literal[str], Literal[api.Message], Literal[api.AgentQuery]],
```

fixieai/agents/agent_func.py
```diff
@@ -0,0 +1,226 @@
...
+            for param in params.values()
+        ):
+            raise TypeError(
+                f"Registered function {func_name} cannot accept variable args: {params!r}."
+            )
+
+        # Ensure every parameter is used exactly once.
+        unresolved_parameters = set(name for name in params.keys())
+
+        # Bind a query/message mapper.
```
hessamb, line 66: Please write a comment paragraph explaining what's going to happen in the rest of this method! It takes a bit time to understand it, especially to someone without context.

hessamb (approved):
Sorry for the delay on this! The changes look good. The new logic for inspecting functions and passing arguments is a bit hard to digest. I don't have a structural change request. But I'd like to ask you to please add comments to make one reader understand the logic without reading the code. And please test fixie-example agents deploy.

PR #158, opened 2023-04-21
Fix `fixie deploy`/`fixie serve --venv` when using relative paths

petersalas:
`subprocess.run` changes the directory before running the executable, so we need to make sure the executable path is made absolute.


petersalas:
(This is why `fixie deploy agents/x --validate` is failing in the fixie-examples repo.)

mdwelsh (approved):

PR #166, opened 2023-05-02
Add markdown support to console.

mdwelsh:
This PR causes output from the console to be rendered as markdown.


NickHeiner:
This is a simple PR, and is fine.

But long-term, I think we should conserve our limited resources by providing functionality in just one channel (e.g. web vs CLI tool). I think we should remove `console` from the CLI tool and push people towards web. (Maybe `console` just becomes an `open https://app.fixie.ai/....` command.)

mdwelsh:
You can pry my text console from my cold dead hands!!! :-)

I agree we want to avoid spending too much time maintaining feature parity here, but I really like using the CLI for debugging.

petersalas (approved):
+1 to @NickHeiner's point, we should assess how worthwhile it is to support a TUI going forward. (Here's [approximate usage](https://app.datadoghq.com/apm/traces?query=resource_name%3A%2AGetSessionByHandle%2A%20OR%20resource_name%3A%2AgetMessages%2A&agg_q=resource_name&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code&historicalData=true&sort_m=&sort_t=&spanType=all&top_n=10&top_o=top&viz=timeseries&x_missing=true&start=1681836332905&end=1683045932905&paused=false) -- `getMessages` is the CLI, `GetSessionByHandle` is the web UI.)

PR #169, opened 2023-05-04
Pin old urllib3 version

petersalas:
`gql[requests]` uses an old version of `requests-toolbelt` which is incompatible with newer versions of `urllib3` (but does not constrain its dependency on `urllib3`).

This adds a version constraint to ensure that `urllib3` is compatible.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -116,7 +116,7 @@ test = ["flake8 (==3.7.8)", "hypothesis (==3.55.3)"]
 
 [[package]]
 name = "cryptography"
-version = "40.0.1"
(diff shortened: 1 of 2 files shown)


juberti (commented):
lgtm, but this seems like a problem for the future, no?

petersalas:
> lgtm, but this seems like a problem for the future, no?

The way I think it will play out is that one of our deps will eventually require `urllib3>=2` and version-solving will fail, and by then hopefully `gql` will allow for newer versions of `requests-toolbelt` that support `urllib3`.

PR #172, opened 2023-05-06
Add docstrings for DocumentCorpus

juberti:


fixieai/agents/corpora.py
```diff
@@ -14,7 +14,10 @@ class DocumentLoader:
 class DocumentCorpus:
     """Document corpus for a Fixie CodeShot Agent."""
 
+    """URLs to load documents from. A trailing wildcard (e.g., https://example.com/*),
```
petersalas, line 17: I think docstrings go below the field they describe (but check the generated docs to be sure)

petersalas (approved):

PR #175, opened 2023-05-11
Support error message propagation from agents

petersalas:
This adds support for propagating `AgentError`s back to Fixie.

It defines an `AgentException` type that is automatically converted into an `AgentResponse` with a FastAPI exception handler, and the func invoker remaps unhandled exceptions to an `AgentException` to handle everything else.

Additionally, `UserStorage.__init__` will now raise an `AgentException` if it is instantiated on behalf of an anonymous user -- that type is refactored to take `VerifiedTokenClaims` to support this.


mdwelsh (approved):
Nice!

PR #178, opened 2023-05-22
Add CorpusRequest and CorpusResponse to fixie-sdk.

mdepinet:
Depends on https://github.com/fixie-ai/fixie/pull/1380


mdepinet:
We're going with #180 instead

PR #180, opened 2023-05-23
Introduce corpus functions

mdepinet:
This is an alternative to https://github.com/fixie-ai/fixie-sdk/pull/178

### DIFF
--- a/fixieai/agents/agent_base.py
+++ b/fixieai/agents/agent_base.py
@@ -7,7 +7,7 @@ import json
 import os
 import re
 import warnings
-from typing import Callable, Dict, Optional
+from typing import Callable, Dict, Optional, TypeVar
 
 import fastapi
 import fastapi.security
@@ -120,23 +120,18 @@ class AgentBase(abc.ABC):
         if func is None:
             # Func is not passed in. It's the decorator being created.
             return functools.partial(self.register_func, func_name=func_name)
+        non_null_func: Callable = func
 
-        if func_name is not None:
-            if not ACCEPTED_FUNC_NAMES.fullmatch(func_name):
-                raise ValueError(
-                    f"Function names may only be alphanumerics, got {func_name!r}."
-                )
-        else:
-            func_name = func.__name__
-
-        if func_name in self._funcs:
-            raise ValueError(f"Func[{func_name}] is already registered with agent.")
-
-        self._funcs[func_name] = agent_func.AgentQueryFunc.create(
+        AgentBase._register_func_internal(
             func,
-            self.oauth_params,
-            default_message_type=api.Message,
-            allow_generator=False,
+            func_name,
+            self._funcs,
+            lambda: agent_func.AgentQueryFunc.create(
+                non_null_func,
+                self.oauth_params,
+                default_message_type=api.Message,
+                allow_generator=False,
+            ),
         )
         return func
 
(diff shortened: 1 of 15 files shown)


fixieai/agents/agent_func.py
```diff
@@ -232,26 +307,93 @@ def create(
...
+        ):
+            raise TypeError(
+                f"Registered function {func_name} cannot accept variable args: {params!r}."
+            )
+
+        if len(params) > 1:
+            raise TypeError(
+                f"Registered function {func_name} must have one or zero arguments."
+            )
+        param = next(iter(params.keys())) if params else None
```
petersalas, line 369:
```suggestion
        param = next(iter(params.keys()), None)
```

fixieai/agents/agent_func.py
```diff
@@ -232,26 +307,93 @@ def create(
...
+                [
+                    (corpora.CorpusRequest, lambda request, claims: request),
+                    (param, lambda request, claims: request),
+                ],
+            )
+            if param
+            else None
+        )
+
+        if bool(params) != bool(bound_request_mapper):
```
petersalas, line 383: A perhaps-simpler alternative to going through the `_bind_argument_mapper` machinery could be to directly validate the type annotation of the parameter, but consistency with the other function is fine too.

fixieai/agents/corpora.py
```diff
@@ -1,18 +1,147 @@
...
 from typing import List, Optional
 
+import dataclasses_json.core
 from pydantic import dataclasses as pydantic_dataclasses
 
 
 @pydantic_dataclasses.dataclass
-class DocumentLoader:
-    """Document loader for a Fixie CodeShot Agent."""
+class CorpusRequest(dataclasses_json.DataClassJsonMixin):
```
petersalas, line 10: Removing `DocumentLoader` seems like a breaking change? Would recommend leaving it for now unless we need to remove it.

fixieai/agents/corpora.py
```diff
@@ -22,4 +151,6 @@ class DocumentCorpus:
     """A list of wildcard patterns to exclude from crawled URLs (e.g., */no_crawl/*)."""
 
     auth_token_func: Optional[str] = None
-    loader: Optional[DocumentLoader] = None
+
+
+DocumentCorpus = UrlDocumentCorpus  # Deprecated: Prefer UrlDocumentCorpus
```
petersalas, line 156: Is there a way to specify the new corpus type?

fixieai/agents/agent_func.py
```diff
@@ -232,26 +307,93 @@ def create(
...
+        if isinstance(result, CorpusResponse):
+            yield result
+        else:
+            raise TypeError(
+                f"The func result was type {type(result)}, but must be a CorpusResponse."
+            )
+
+    def adapt_exception(self, e: exceptions.AgentException) -> CorpusResponse:
+        logging.warn(f"Corpus loading threw an exception: {e}", stack_info=True)
+        return corpora.CorpusResponse()
```
petersalas, line 335: I think this should probably just raise and HTTP 500 -- unlike `AgentQuery`/`AgentReponse` there's not a human or LLM on the other end of the request.

fixieai/agents/agent_base.py
```diff
@@ -195,6 +251,45 @@ def _serve_func(
...
+                http_status_code=404,
+                func_name=corpus_func_name,
+                available_funcs=list(self._corpus_funcs.keys()),
+            )
+
+        output = pyfunc(request, token_claims)
+
+        return next(iter(output))
+
+    def _check_credentials(
```
petersalas, line 283: might as well use this in `standalone.py` as well

petersalas (approved):
I like it! Curious what @juberti's take is as well.

fixieai/agents/corpora.py
mdepinet: AFAICT, this never did anything so I figured we could bump the version to be safe and remove it.  For now I've added it back and deprecated it instead.

fixieai/agents/corpora.py
mdepinet: Whoops, I forgot to add the registered function to the agent metadata. 😅 Added.

fixieai/agents/agent_func.py
mdepinet: I initially wasn't sure this was going to be simpler, but I think you're right. Done.

PR #182, opened 2023-05-24
Change `fixie serve` and `fixie deploy` to use revisions API

petersalas:
This adds revisions-related commands to the CLI and methods to `FixieClient`. Notably:
- `fixie agent revisions list` lists all revisions associated with an agent
- `fixie agent revisions set-current` sets the current revision
- `fixie agent revisions delete` deletes a revision

Additionally, this changes the `fixie agent serve` and `fixie agent deploy` commands to use the revisions API. Functionally this means:
- `fixie serve` restores the pre-existing revision when exiting and deletes ephemeral revisions
- `fixie deploy` now supports corpora reindexing, deploying a non-current revision, and tagging revisions with user-provided metadata

From an implementation standpoint, this requires removing a dependency on the "refresh-on-startup" hook so that the CLI can actively manage (and dispose of) the ephemeral revisions. The CLI now
- polls the host/port to see when the server has started up
- handles server reloads itself rather than delegating it to `uvicorn`

### DIFF
--- a/fixieai/cli/agent/commands_test.py
+++ b/fixieai/cli/agent/commands_test.py
@@ -1,9 +1,12 @@
 import os
 import pathlib
+import queue
 import random
 import string
-import subprocess
-from unittest import mock
+import threading
+import time
+import unittest.mock
+from typing import Callable, Dict, Optional
 
 import click.testing
 import gql.client
@@ -14,7 +17,6 @@ import fixieai
 from fixieai import constants
 from fixieai.cli import cli
 from fixieai.cli.agent import agent_config
-from fixieai.client.client import FixieClient
 
 from . import commands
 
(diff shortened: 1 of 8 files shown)


fixieai/cli/agent/commands_test.py
```diff
@@ -176,45 +248,211 @@ def test_init_validation_prompting():
         assert config.more_info_url == ""
 
 
+def test_temporary_revision_replacer(mock_graphql):
```
mdwelsh, line 251: Scratching my head about what you're doing here. Mind adding some comments?

fixieai/client/client.py
```diff
@@ -211,6 +212,140 @@ def refresh_agent(self, agent_handle: str):
...
+        self,
+        handle: str,
+        make_current: bool,
+        *,
+        reindex_corpora: bool = False,
+        metadata: Optional[Dict[str, str]] = None,
+        external_url: Optional[str] = None,
+        python_gzip_tarfile: Optional[BinaryIO] = None,
+    ) -> str:
+        """Creates a new Agent revision."""
```
mdwelsh, line 225: Would you mind adding "Args:" to the docstring explaining all of the exciting options?

mdwelsh (approved):
Nice! A couple of nitpicks.

fixieai/cli/agent/commands_test.py
petersalas: Ha, fair enough :)

fixieai/client/client.py
petersalas: Good call, added.

PR #192, opened 2023-08-01
Allow `fixie agent deploy` to deploy node.js agents

juberti:
The environment decision is triggered by the presence or absence of `package.json` in the root agent dir.

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -598,7 +598,7 @@ def deploy(ctx, path, metadata_only, public, validate):
                         tarball,
                     )
                     tarball_file.seek(0)
-                    with _spinner(console, "Deploying..."):
+                    with _spinner(console, "Deploying Python..."):
                         ctx.obj.client.deploy_agent(
                             config.handle,
                             tarball_file,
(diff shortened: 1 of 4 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -711,6 +713,21 @@ def _agent_code_package(agent_dir: str, agent_id: str, console: rich_console.Con
...
+    with tempfile.TemporaryDirectory() as tarball_dir:
+        package_json_path = os.path.join(agent_path, "package.json")
+        with open(package_json_path) as package_json_file:
+            package_json = json.load(package_json_file)
+        tarball_path = os.path.join(
+            tarball_dir, f"{package_json['name']}-{package_json['version']}.tgz"
+        )
+        subprocess.check_call(["npm", "pack", agent_path], cwd=tarball_dir)
+        tarball_file = open(tarball_path, "rb")
+    yield tarball_file
```
petersalas, line 728: Is it safe (e.g. on windows?) to `yield` outside of the `with`?

fixieai/client/client.py
```diff
@@ -221,7 +221,8 @@ def create_agent_revision(
         reindex_corpora: bool = False,
         metadata: Optional[Dict[str, str]] = None,
         external_url: Optional[str] = None,
-        python_gzip_tarfile: Optional[BinaryIO] = None,
+        gzip_tarfile: Optional[BinaryIO] = None,
+        environment: str = "PYTHON",
```
mdwelsh, line 225: NIT: It would be better to use an Enum for this field.

mdwelsh (approved):
Nice!

petersalas (approved):
LGTM! Are we able to test e2e or is there more platform work to do before then?

PR #195, opened 2023-08-03
Initial JS support for `fixie agent serve`

juberti:
LMK if you have thoughts on how to install packages in a venv-like manner...

### DIFF
--- a/fixieai/cli/agent/commands.py
+++ b/fixieai/cli/agent/commands.py
@@ -270,6 +270,13 @@ def delete_agent(ctx, handle: str):
     click.echo(f"Deleted agent {agent.agent_id}")
 
 
+def _get_agent_type(agent_dir) -> fixieai.client.FixieEnvironment:
+    if os.path.exists(os.path.join(agent_dir, "package.json")):
+        return fixieai.client.FixieEnvironment.NODEJS
+    else:
+        return fixieai.client.FixieEnvironment.PYTHON
+
+
 def _validate_agent_path(ctx, param, value):
     normalized = agent_config.normalize_path(value)
     if not os.path.exists(normalized):
@@ -427,7 +434,8 @@ def _temporary_revision_replacer(
 def _server_process(
     console: rich_console.Console,
     on_exit_event: threading.Event,
-    python_exe: str,
+    agent_type: fixieai.client.FixieEnvironment,
+    agent_exe: str,
     host: str,
     port: int,
     agent_env: Dict[str, str],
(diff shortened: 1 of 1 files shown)


fixieai/cli/agent/commands.py
```diff
@@ -531,12 +549,21 @@ def serve(ctx, path, host, port, use_tunnel, reload, use_venv):
...
-        else:
-            python_exe, agent_env = sys.executable, os.environ.copy()
+        if agent_type == fixieai.client.FixieEnvironment.PYTHON:
+            if use_venv:
+                agent_exe, agent_env = _configure_venv(console, agent_dir)
+            else:
+                agent_exe, agent_env = sys.executable, os.environ.copy()
+        elif agent_type == fixieai.client.FixieEnvironment.NODEJS:
+            if use_venv:
+                # TODO: Implement virtual environment for node
```
mdwelsh, line 562: I think the "venv" is actually the `node_packages` directory that gets created when you do an `npm install` -- I guess the concern here is you could end up clobbering whatever is already there but this should not be a concern since I think people using `fixie agent serve` will expect it to use the version here. In other words I don't think you *need* anything special for a "venv"...

mdwelsh (approved):

fixieai/cli/agent/commands.py
```diff
@@ -447,7 +455,17 @@ def _server_process(
             "--port",
             str(port),
             "--factory",
-        ],
+        ]
+    elif agent_type == fixieai.client.FixieEnvironment.NODEJS:
+        args = [
+            agent_exe,
+            "ts-node",
```
petersalas, line 462: I saw this on build PR as well -- why do we need to use `ts-node`? Don't have a particular problem with it, just seems odd since we're ostensibly executing a `.js` file.

fixieai/cli/agent/commands.py
petersalas: Yeah I think the equivalent would be the CLI _doing_ the `npm install` -- and like Matt says that would already be local to the `node_modules` directory.

I think it's probably reasonable to assume that the user has already done the `npm install`.

petersalas (approved):

fixieai/cli/agent/commands.py
juberti: I just copied this from the previous SDK branch. Should I just run "node" here?

fixieai/cli/agent/commands.py
juberti: Thanks, that makes sense. I'll cull this for now and we can revisit later if needed.

PR #196, opened 2023-08-03
Add token streaming support to CLI

juberti:
This switches the client over to use the REST /api/agents API rather than the graphql API, and accordingly allows us to remove some of the oddness of the old usage, including the need for a separate thread and polling of the get_messages API.

### DIFF
--- a/fixieai/client/client.py
+++ b/fixieai/client/client.py
@@ -3,8 +3,9 @@
 from __future__ import annotations
 
 import enum
+import json
 import logging
-from typing import Any, BinaryIO, Dict, List, Optional
+from typing import Any, BinaryIO, Dict, Generator, List, Optional
 
 import requests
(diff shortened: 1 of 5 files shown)


fixieai/client/client.py
```diff
@@ -382,3 +377,17 @@ def deploy_agent(
             gzip_tarfile=gzip_tarfile,
             environment=environment,
         )
+
+    def post(self, url: str, data: Optional[Dict[str, Any]] = None) -> Any:
```
mdwelsh, line 381: Since these APIs are part of our SDK documentation, do you mind adding docstrings here? Also for streaming_post.

mdwelsh (approved):
Nice!

petersalas (approved):
LGTM!

PR #202, opened 2023-11-30
forklift

benlower:
replacing this with the JS/TS SDK and CLI


juberti:
if all the JS stuff is going to be in the root, we should call this fixie-sdk-js. Otherwise we should do a monorepo with different dirs for each language.

juberti (approved):
approving, we can rename the repo at any time

PR #207, opened 2023-12-07
Some changes to logging and initialization logic.

mdwelsh:
These changes are meant to address a couple of startup issues:

* There is a separate 'startAudio()' step that can be optionally called before 'start()' to get mic permissions before the peer starts speaking.
* Added a 'audioContext.resume()' in '.startAudio()' to fix issues where the AudioContext was initialized outside of a click handler and needed to be resumed.
* Added a bit of extra logging to help with debugging.
* There are some minor style cleanups here mostly to make lint happy.

PR #208, opened 2023-12-08
Fix race condition with track subscriptions

juberti:
- Ensures our event listeners are installed before we connect to the room to ensure we don't miss any events.
- Also includes a followup to #207, specifically https://github.com/fixie-ai/fixie-sdk-js/pull/207#discussion_r1419745465

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,4 +1,4 @@
-nodeLinker: pnp
+nodeLinker: node-modules
 
 packageExtensions:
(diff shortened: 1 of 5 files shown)


mdwelsh (commented):
I responded to this. These need to be here because you call .warmup after the socket closes. If you don't have these here then it still thinks it's started, which is not correct.

mdwelsh (approved):

PR #210, opened 2023-12-10
Improved error handling in the voice client.

mdwelsh:
This PR makes a few minor changes to improve error handling and debugging.

- We explicitly track the initializing state of the VoiceSession as it goes from created -> connecting -> connected -> audio enabled -> started. This is going to make it easier for applications to observe the state of the session and take appropriate action when errors or timeouts occur.
- The onError handler carries a VoiceSessionError object with information about the error. This is a simple message for now, but can be expanded later.
- Added a few additional places where .onError is called.

The more substantive (perhaps) change is that .warmup is no longer called in the ctor or when the socket closes. I would much rather this be done explicitly by application code, rather than the VoiceSession trying to be too clever. Constructors are not a good place to be doing actions like opening socket connections, anyway, because it's much harder to deal with errors that happen during the ctor. By having .warmup() be called explicitly, we can create the VoiceSession, set up the error handling logic, call .warmup, and deal gracefully with connection errors that might occur.

These changes allow us to have more graceful error handling in the Santa app.

### DIFF
--- a/packages/fixie/src/voice.ts
+++ b/packages/fixie/src/voice.ts
@@ -89,9 +89,6 @@ export class VoiceSession {
   /** Called when an error occurs. */
   onError?: (error: VoiceSessionError) => void;
 
-  /** Called when audio starts. */
-  onAudioStart?: () => void;
-
   constructor(
     private readonly agentId: AgentId,
     public conversationId?: ConversationId,
@@ -148,7 +145,6 @@ export class VoiceSession {
       this.sendData(obj);
     }, 5000);
     this.audioStarted = true;
-    this.onAudioStart?.();
   }
 
   /** Start this VoiceSession. This will call startAudio if it has not been called yet.*/
--- a/packages/fixie/src/voice.ts
+++ b/packages/fixie/src/voice.ts
@@ -89,11 +89,7 @@ export class VoiceSession {
   /** Called when an error occurs. */
   onError?: (error: VoiceSessionError) => void;
 
-  constructor(
-    private readonly agentId: AgentId,
-    public conversationId?: ConversationId,
-    private readonly params?: VoiceSessionInit
-  ) {
+  constructor(readonly agentId: AgentId, public conversationId?: ConversationId, readonly params?: VoiceSessionInit) {
     console.log('[voiceSession] creating VoiceSession');
   }
 
--- a/packages/fixie/src/voice.ts
+++ b/packages/fixie/src/voice.ts
@@ -163,7 +163,6 @@ export class VoiceSession {
     }
     this.started = true;
     this.maybePublishLocalAudio();
-    this.changeState(VoiceSessionState.IDLE);
   }
 
   /** Stop this VoiceSession. */
@@ -210,6 +209,7 @@ export class VoiceSession {
       console.log('[voiceSession] publishing local audio track');
       const opts = { name: 'audio', simulcast: false, source: Track.Source.Microphone };
       this.room.localParticipant.publishTrack(this.localAudioTrack, opts);
+      this.changeState(VoiceSessionState.IDLE);
     } else {
       console.log(
         `[voiceSession] not publishing local audio track - room state is ${this.room?.state}, local audio is ${
--- a/packages/fixie/src/voice.ts
+++ b/packages/fixie/src/voice.ts
@@ -94,8 +94,8 @@ export class VoiceSession {
 
   constructor(
     private readonly agentId: AgentId,
-    private readonly conversationId?: ConversationId,
-    private readonly params?: VoiceSessionInit
+    private readonly params?: VoiceSessionInit,
+    public conversationId?: ConversationId,
   ) {
     console.log('[voiceSession] creating VoiceSession');
   }
@@ -214,7 +214,7 @@ export class VoiceSession {
       console.log(
         `[voiceSession] not publishing local audio track - room state is ${this.room?.state}, local audio is ${
           this.localAudioTrack != null
-        }`
+        }`,
       );
     }
   }
@@ -258,7 +258,7 @@ export class VoiceSession {
         this.maybePublishLocalAudio();
         this.room.on(RoomEvent.TrackSubscribed, (track: RemoteTrack) => this.handleTrackSubscribed(track));
         this.room.on(RoomEvent.DataReceived, (payload: Uint8Array, participant: any) =>
-          this.handleDataReceived(payload, participant)
+          this.handleDataReceived(payload, participant),
         );
         break;
       default:
@@ -318,6 +318,8 @@ export class VoiceSession {
       this.handleOutputChange(msg.text, msg.final);
     } else if (msg.type == 'latency') {
       this.handleLatency(msg.kind, msg.value);
+    } else if (msg.type == 'conversation_created') {
+      this.conversationId = msg.conversationId;
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/fixie/src/voice.ts
```diff
@@ -253,25 +268,25 @@ export class VoiceSession {
...
   }
 
   private handleTrackSubscribed(track: RemoteTrack) {
     console.log(`[voiceSession] subscribed to remote audio track ${track.sid}`);
     const audioTrack = track as RemoteAudioTrack;
     audioTrack.on(TrackEvent.AudioPlaybackStarted, () => console.log('[voiceSession] audio playback started'));
-    audioTrack.on(TrackEvent.AudioPlaybackFailed, (err: any) =>
-      console.error('[voiceSession] audio playback failed', err)
-    );
+    audioTrack.on(TrackEvent.AudioPlaybackFailed, (err: any) => {
```
juberti, line 286: this shouldn't be emitted, this is an internal detail and not actionable by the application

packages/fixie/src/voice.ts
```diff
@@ -11,10 +13,12 @@ import {
...
 
-/**
- * Represents the state of a VoiceSession.
- */
+/** Represents the state of a VoiceSession. */
 export enum VoiceSessionState {
+  DISCONNECTED = 'disconnected',
+  CONNECTING = 'connecting',
+  CONNECTED = 'connected',
+  AUDIO_STARTED = 'audio_started',
```
juberti, line 21: these states are not linear, e.g., you can be AUDIO_STARTED without being CONNECTED. The overall approach makes sense, but I think some refinement of the states is needed.

packages/fixie/src/voice.ts
```diff
@@ -110,13 +116,20 @@ export class VoiceSession {
...
-    this.socket.onopen = () => this.handleSocketOpen();
-    this.socket.onmessage = (event) => this.handleSocketMessage(event);
-    this.socket.onclose = (event) => this.handleSocketClose(event);
+    try {
+      const url = this.params?.webrtcUrl ?? 'wss://wsapi.fixie.ai';
+      this.socket = new WebSocket(url);
+      this.socket.onopen = () => this.handleSocketOpen();
+      this.socket.onmessage = (event) => this.handleSocketMessage(event);
+      this.socket.onclose = (event) => this.handleSocketClose(event);
+      this.changeState(VoiceSessionState.CONNECTED);
```
juberti, line 128: this is the wrong time to go to CONNECTED, doesn't occur until onopen

packages/fixie/src/voice.ts
mdwelsh: Right - makes sense. I'll add a separate event for that.

packages/fixie/src/voice.ts
mdwelsh: Gotcha, thanks.

packages/fixie/src/voice.ts
mdwelsh: Good catch - thanks

mdwelsh:
Thanks! PTAL

packages/fixie/src/voice.ts
```diff
@@ -147,6 +163,7 @@ export class VoiceSession {
     }
     this.started = true;
     this.maybePublishLocalAudio();
+    this.changeState(VoiceSessionState.IDLE);
```
juberti, line 166: the socket might not yet be connected here, I think this needs to happen in maybePublishLocalAudio

packages/fixie/src/voice.ts
```diff
@@ -110,13 +118,20 @@ export class VoiceSession {
   /** Warm up the VoiceSession by making network connections. This is called automatically when the VoiceSession object is created. */
   warmup(): void {
     console.log('[voiceSession] warming up');
+    this.changeState(VoiceSessionState.CONNECTING);
```
juberti, line 121: I'd trigger this after we create the socket

mdwelsh: So, that leads to a situation where if you get an error on socket connect, the state of the VoiceSession will not always be in the same state. It might be DISCONNECTED (if the error happens soon after socket creation) or CONNECTING (if the error happens a bit later). I thought it would be better to be consistent so one dealing with errors doesn't need to understand why it might be one state or another.

packages/fixie/src/voice.ts
```diff
@@ -133,6 +148,7 @@ export class VoiceSession {
       this.sendData(obj);
     }, 5000);
     this.audioStarted = true;
+    this.onAudioStart?.();
```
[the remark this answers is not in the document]
mdwelsh: I assumed this would be useful here because if the user declined mic access presumably (I am not sure?) the localAudioTrack would not be populated and then we would not fire this event. The idea was to only fire the event once mic access had been granted and obtained by the app. WDYT?

packages/fixie/src/voice.ts
mdwelsh: Good catch, done

packages/fixie/src/voice.ts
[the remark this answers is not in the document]
juberti: we'll block in this method until we get an answer from the user. If the user says no we'll get a permission-denied `DOMException` which the app can catch when calling `startAudio`.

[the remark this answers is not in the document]
juberti: IOW I still tend to think the event is not valuable here since it's just tracking whether a single method completed.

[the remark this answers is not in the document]
mdwelsh: OK, removed it!

packages/fixie/src/voice.ts
```diff
@@ -112,11 +117,18 @@ export class VoiceSession {
...
-    this.socket.onmessage = (event) => this.handleSocketMessage(event);
-    this.socket.onclose = (event) => this.handleSocketClose(event);
+    try {
+      const url = this.params?.webrtcUrl ?? 'wss://wsapi.fixie.ai';
+      this.socket = new WebSocket(url);
+      this.socket.onopen = () => this.handleSocketOpen();
+      this.socket.onmessage = (event) => this.handleSocketMessage(event);
+      this.socket.onclose = (event) => this.handleSocketClose(event);
+      this.changeState(VoiceSessionState.CONNECTING);
+    } catch (e) {
```
juberti, line 127: can you `catch (e: Error) {`

mdwelsh: I tried this before, but eslint is strict on this for a good reason - there's no guarantee the thing you are catching here is, in fact, an Error object!

[
Screenshot 2023-12-11 at 3 05 05 PM
](url)

packages/fixie/src/voice.ts
```diff
@@ -94,8 +94,8 @@ export class VoiceSession {
 
   constructor(
     private readonly agentId: AgentId,
-    private readonly conversationId?: ConversationId,
-    private readonly params?: VoiceSessionInit
+    private readonly params?: VoiceSessionInit,
+    public conversationId?: ConversationId,
```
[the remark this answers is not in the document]
mdwelsh: Done

packages/fixie/src/voice.ts
juberti: yeah, sorry, forgot JS only lets you have a single catch()

PR #211, opened 2023-12-12
Send event when conversation ID changes.

mdwelsh:
This is necessary for the HiSanta app to pick this value up and include it in metrics.

### DIFF
--- a/packages/fixie/src/voice.ts
+++ b/packages/fixie/src/voice.ts
@@ -89,6 +89,9 @@ export class VoiceSession {
   /** Called when an error occurs. */
   onError?: (error: VoiceSessionError) => void;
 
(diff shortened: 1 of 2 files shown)


juberti (commented):
How does this event get fired?

packages/fixie/src/voice.ts
```diff
@@ -89,6 +89,9 @@ export class VoiceSession {
   /** Called when an error occurs. */
   onError?: (error: VoiceSessionError) => void;
 
+  /** Called when the conversation ID is changed. */
+  onConversationIdChange?: (conversationId: ConversationId) => void;
```
juberti, line 93: suggest calling this `onConversationCreated`

PR #212, opened 2023-12-14
Expose room ID from the voiceSession.

mdwelsh:
This is necessary for recording sharing.


packages/fixie/src/voice.ts
```diff
@@ -108,6 +108,11 @@ export class VoiceSession {
     return this.outAnalyzer?.analyzer;
   }
 
+  /** Returns the Room ID currently in use by this VoiceSession. */
+  get roomId() : string | undefined {
```
juberti, line 112: Technically this is a roomName, livekit rooms also have room IDs but they are different.

juberti (approved):

PR #216, opened 2024-01-04
Refactoring of Fixie JS SDK.

mdwelsh:
This PR refactors the SDK into three packages:

* `@fixieai/fixie-common` - Base package for both NodeJS and browser environments, containing common code for both. Not intended to be used directly.
* `fixie` - NodeJS package, based on top of `@fixieai/fixie-common`, containing code for the CLI, agent deployment, etc.
* `fixie-web`- Browser package, based on top of `@fixieai/fixie-common`, containing code for browser environments, including the Voice Session API.

The goals here are to make it possible to cleanly separate code for the browser and NodeJS environments, manage dependencies in a sane way between the two environments, and avoid code duplication by having a common environment-neutral package.

This PR also adds unit tests for several of the modules to provide some baseline structure for testing and CI going forward.

For current users of the Fixie package in NodeJS environments, everything should work as it currently does. The CLI works identically as well.

Browser-based users will need to migrate to the new `fixie-web` package for voice and React hooks, etc.


README.md
benlower, line 1: I don't want to add complexity but what about the Python version? Can `fixie-common` help there too?

.prettierignore
```diff
@@ -1,7 +1,4 @@
 .pnp*
 .yarn*
 .vscode*
-packages/examples/src/bakeoff/*/data/*
```
benlower, line 4: what was this used for?

README.md
```diff
@@ -1 +1,24 @@
...
+
+This is a monorepo providing a TypeScript / JavaScript SDK to the
+[Fixie](https://fixie.ai) platform. It contains the following packages:
+
+- `fixie`: A NodeJS SDK and CLI.
+- `fixie-web`: A browser-based SDK.
+- `fixie-common`: A shared package containing code used by both the
+  NodeJS and browser SDKs.
+
+Full documentation is provided at [https://fixie.ai/docs](https://fixie.ai/docs).
```
benlower, line 11: could we highlight the differences between web & common here? voice is the only thing right now?

packages/fixie/tests/fixtures/test-fixie-config.yaml
```diff
@@ -0,0 +1,2 @@
+apiUrl: https://fake.api.domain
```
benlower, line 1: curious why we need to have apiUrl duplicated across this file and in `test-fixie-config-ignore-fields.yaml`

packages/fixie/tests/auth.test.ts
```diff
@@ -0,0 +1,53 @@
+/** Unit tests for auth.ts. */
+
+//console.log("MDW INIT auth.test.ts");
```
benlower, line 3: does this need to be here?

README.md
mdwelsh: Python SDK is a different project, way out of scope for this PR.

.prettierignore
mdwelsh: This was some thing that I think Hessam was doing to compare different LLMs or something. Not in this repo anyway so we can remove this.

README.md
mdwelsh: I don't think we want to stuff a lot of details into this README right now, but if you want to flesh this out as part of the documentation refactor please go for it. My concern is generally around duplicating information in multiple places and having to keep it up to date across the README and the "real" documentation. This PR is not really about the documentation, it's for the code structuring.

packages/fixie/tests/fixtures/test-fixie-config.yaml
mdwelsh: They are two different files used in two different tests. One of them tests what happens when the file is nominal, i.e., no bad or unknown fields, the other has the additional unknown field in it.

packages/fixie/tests/auth.test.ts
mdwelsh: Good catch! Removed.

benlower (approved):

PR #223, opened 2024-01-13
Fix agent deployment.

mdwelsh:
This PR fixes a bug with how FixieAgent.createRevision was using the new Agent revisions API. This also fixes agent deployment with the new API.

### DIFF
--- a/packages/fixie/src/agent.ts
+++ b/packages/fixie/src/agent.ts
@@ -188,6 +188,7 @@ export class FixieAgent extends FixieAgentBase {
         displayName: config.name,
         description: config.description,
         moreInfoUrl: config.moreInfoUrl,
+        published: true,
       })) as FixieAgent;
     }
     return agent as FixieAgent;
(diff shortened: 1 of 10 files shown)


packages/fixie/src/agent.ts
```diff
@@ -113,68 +113,42 @@ export class FixieAgent extends FixieAgentBase {
...
-   *   May only be specified if `externalUrl` is also specified.
    * @param isCurrent Whether this revision should be the current revision.
-   * @param tarball The path to the tarball containing the agent code.
    * @param environmentVariables The environment variables to set for the agent.
-   *   May only be specified if `tarball` is also specified.
    *
    * @returns The created agent revision.
    */
-  public async createRevision({
+  public async createManagedRevision({
```
petersalas, line 126: How does `fixie serve` work without the `externalUrl` support?

mdwelsh: It doesn't use this code path. It uses the base class `createRevision` instead. See line 431.

petersalas: I see -- I don't really understand why the code is split across the two packages the way it is, but presumably you have a reason :)

petersalas (approved):

PR #224, opened 2024-01-18
Auto-warmup VoiceSession on start()

juberti:


packages/fixie-web/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fixie-web",
   "description": "Browser-based SDK for the Fixie platform.",
-  "version": "1.0.8",
+  "version": "1.0.9",
```
mdwelsh, line 4: Would you mind using the new-and-improved way of doing this for this PR? Here's what you do:

1. Run `yarn changeset` at the root
2. It will ask which packages need a version bump at at what level (patch is fine)
3. Run `yarn changeset version` which will do the actual version bump and prompt you for a changelist description
4. Commit.

This may be overkill for a little PR like this but I'm trying to get us in the habit of managing dependencies across the different package versions in a sane way. 

After the PR lands, you can then do `yarn changeset publish --tag unstable` to publish all of the changes packages in the tree to npm (I will do this since I'm working on a change that needs to land as well).

mdwelsh (approved):

packages/fixie-web/package.json
juberti: Will do when back at desk

juberti: maybe I'm not doing things right, but I'm getting
```
(base) juberti@juberti-macbook-pro fixie-sdk % yarn changeset  
Usage Error: Couldn't find a script named "changeset".
```

PR #226, opened 2024-01-19
Fix issue with 204 responses raising an error.

mdwelsh:
204 responses (e.g., from deleteRevision) were raising errors, this fixes that.

### DIFF
--- a/packages/fixie-common/src/agent.ts
+++ b/packages/fixie-common/src/agent.ts
@@ -277,7 +277,6 @@ export class FixieAgentBase {
     if (externalUrl !== undefined) {
       externalDeployment = {
         url: externalUrl,
-        runtimeParametersSchema,
       };
     }
     const result = (await this.client.requestJson(`/api/v1/agents/${this.metadata.agentId}/revisions`, {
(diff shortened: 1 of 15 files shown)


packages/fixie-common/src/agent.ts
```diff
@@ -301,11 +301,7 @@ export class FixieAgentBase {
...
   }
 
-  public deleteRevision(revisionId: string): Promise<void> {
-    return this.client.requestJson(
-      `/api/v1/agents/${this.metadata.agentId}/revisions/${revisionId}`,
-      undefined,
-      'DELETE'
-    );
+  public deleteRevision(revisionId: string) {
+    this.client.requestJson(`/api/v1/agents/${this.metadata.agentId}/revisions/${revisionId}`, undefined, 'DELETE');
```
petersalas, line 305: - Should either `return` or `await` the promise
- Should this just be calling `this.client.request`? Seems weird to invoke `requestJson` when there's no JSON

packages/fixie-common/src/client.ts
```diff
@@ -100,6 +100,9 @@ export class FixieClientBase {
 
   async requestJson<T = Jsonifiable>(path: string, bodyData?: unknown, method?: string): Promise<T> {
     const response = await this.request(path, bodyData, method);
+    if (response.status === 204) {
```
petersalas, line 103: Should this be checking `Content-Type` instead?

mdwelsh: I don't think so. 204 has no content-type, right?

packages/fixie-common/src/agent.ts
mdwelsh: Yeah, I'll clean this up everywhere.

mdwelsh (commented):
Thanks - PTAL

petersalas (approved):

PR #227, opened 2024-01-22
Add support for specifying default parameters on CLI

petersalas:
This ports https://github.com/fixie-ai/ai-jsx/pull/504 over to this repo. This also changes the function signature of `createRevision` to take JSON objects and uses those objects for the underlying API calls.

### DIFF
--- a/packages/fixie-common/src/agent.ts
+++ b/packages/fixie-common/src/agent.ts
@@ -264,7 +264,7 @@ export class FixieAgentBase {
   }: {
     defaultRuntimeParameters?: Record<string, unknown>;
     externalUrl?: string;
-    runtimeParametersSchema?: string;
+    runtimeParametersSchema?: Record<string, unknown> | null;
     isCurrent?: boolean;
   }): Promise<AgentRevision> {
     if (externalUrl === undefined && defaultRuntimeParameters === undefined) {
@@ -287,10 +287,9 @@ export class FixieAgentBase {
           external: externalDeployment,
         },
         runtime: {
-          parametersSchema: runtimeParametersSchema,
+          parametersSchema: runtimeParametersSchema && JSON.stringify(runtimeParametersSchema),
         },
-        // The API expects this field to be pre-JSON-encoded.
-        defaultRuntimeParameters: JSON.stringify(defaultRuntimeParameters),
+        defaultRuntimeParameters: defaultRuntimeParameters && JSON.stringify(defaultRuntimeParameters),
       },
     })) as { revision: AgentRevision };
     return result.revision;
--- a/packages/fixie-common/src/agent.ts
+++ b/packages/fixie-common/src/agent.ts
@@ -264,7 +264,7 @@ export class FixieAgentBase {
   }: {
     defaultRuntimeParameters?: Record<string, unknown>;
     externalUrl?: string;
-    runtimeParametersSchema?: Record<string, unknown> | null;
+    runtimeParametersSchema?: Record<string, unknown>;
     isCurrent?: boolean;
   }): Promise<AgentRevision> {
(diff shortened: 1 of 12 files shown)


packages/fixie-common/src/agent.ts
```diff
@@ -287,10 +287,9 @@ export class FixieAgentBase {
           external: externalDeployment,
         },
         runtime: {
-          parametersSchema: runtimeParametersSchema,
+          parametersSchema: runtimeParametersSchema && JSON.stringify(runtimeParametersSchema),
```
mdwelsh, line 290: Is this the right thing? If you `JSON.stringify` a string, doesn't that wrap it in extra quotes we don't want here?

packages/fixie-common/src/agent.ts
```diff
@@ -287,10 +287,9 @@ export class FixieAgentBase {
...
         },
-        // The API expects this field to be pre-JSON-encoded.
-        defaultRuntimeParameters: JSON.stringify(defaultRuntimeParameters),
+        defaultRuntimeParameters: defaultRuntimeParameters && JSON.stringify(defaultRuntimeParameters),
```
mdwelsh, line 292: I'm confused here, why would you need the test?

packages/fixie-common/src/agent.ts
```diff
@@ -264,7 +264,7 @@ export class FixieAgentBase {
   }: {
     defaultRuntimeParameters?: Record<string, unknown>;
     externalUrl?: string;
-    runtimeParametersSchema?: string;
+    runtimeParametersSchema?: Record<string, unknown> | null;
```
mdwelsh, line 267: Why can this be `null` ?

packages/fixie/src/agent.ts
```diff
@@ -87,12 +87,7 @@ export class FixieAgent extends FixieAgentBase {
       `
     );
     const program = TJS.programFromConfig(tsconfigPath, [tempPath]);
-    const schema = TJS.generateSchema(program, 'RuntimeParameters', settings);
-    if (schema && schema.type !== 'object') {
-      throw new Error(`The first argument of your default export must be an object (not ${schema.type})`);
-    }
-
-    return schema;
+    return TJS.generateSchema(program, 'RuntimeParameters', settings);
```
mdwelsh, line 90: Is this check not still relevant?

packages/fixie/src/agent.ts
```diff
@@ -132,7 +127,7 @@ export class FixieAgent extends FixieAgentBase {
   }: {
     tarball: string;
     defaultRuntimeParameters?: Record<string, unknown>;
-    runtimeParametersSchema?: string;
+    runtimeParametersSchema?: Record<string, unknown> | null;
```
mdwelsh, line 130: How can this be null?

packages/fixie/src/agent.ts
```diff
@@ -143,15 +138,15 @@ export class FixieAgent extends FixieAgentBase {
       revision: {
         isCurrent,
         runtime: {
-          parametersSchema: runtimeParametersSchema,
+          parametersSchema: runtimeParametersSchema && JSON.stringify(runtimeParametersSchema),
```
mdwelsh, line 141: Same comments as elsewhere, I don't understand these `&&` checks

packages/fixie-common/src/agent.ts
petersalas: It was/remains optional; I consolidated with `undefined` instead.

packages/fixie/src/agent.ts
petersalas: It's overly restrictive and precludes `oneOf`/`allOf`. The actual requirement is that an empty object satisfies the schema, which the service verifies.

mdwelsh (approved):

PR #228, opened 2024-01-25
initial version

benlower:
i want to refactor this to use the fixie SDK after it supports list agent conversations.


packages/examples/usage-tracking/usage-tracking.js
```diff
@@ -0,0 +1,106 @@
...
+    // Iterate through the turn messages and log their stats
+    turn.messages.forEach((message) => {
+      // console.log(message);
+      numMessages++;
+      // Make sure it's a message type that we want
+      if (turn.state == 'done') {
+        // Function Response
+        if (turn.role == 'assistant' && message.kind == 'functionResponse') {
+          lenConversations += message.response.length;
+          const encoded = encode(message.response);
```
mdwelsh, line 69: I don't know for sure that this is the right encoder, it might be though. The one we use in Python land is 

```
import tiktoken

# OpenAI models that we care about (GPT-3.5, GPT-4, text-embedding-ada-002) use cl100k_base.
#   https://github.com/openai/tiktoken/blob/main/tiktoken/model.py
#   https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
encoder = tiktoken.get_encoding("cl100k_base")
```

mdwelsh (approved):
LGTM for now, run this by @petersalas though

packages/examples/usage-tracking/usage-tracking.js
```diff
@@ -0,0 +1,106 @@
...
+  console.log(`Total conversation characters\t${numConvoChars}`);
+  console.log(`Total conversation tokens\t${numConvoTokens}`);
+}
+
+async function getAgentConversations(agentId) {
+  try {
+    const response = await axios({
+      method: 'get',
+      maxBodyLength: Infinity,
+      url: `https://api.fixie.ai/api/v1/agents/${agentId}/conversations`,
```
petersalas, line 96: Is this endpoint paginated?

packages/examples/usage-tracking/usage-tracking.js
```diff
@@ -0,0 +1,106 @@
...
+      // Make sure it's a message type that we want
+      if (turn.state == 'done') {
+        // Function Response
+        if (turn.role == 'assistant' && message.kind == 'functionResponse') {
+          lenConversations += message.response.length;
+          const encoded = encode(message.response);
+          numTokens += encoded.length;
+
+          numConvoChars += message.response.length;
+          numConvoTokens += encoded.length;
```
petersalas, line 73: This ignores the quadratic effect where messages become input tokens to subsequent generations (e.g. a conversation with 10 messages uses >> 10 times the number of input tokens as a conversation with 1 message)

But old messages also get pruned as well as they fill up the context window, so calculating it accurately is even trickier. The easiest thing to get approximately correct is to calculate the number of _output_ tokens, because that's just the assistant text/functionCall messages, but I'm guessing you want an estimate for both :)

packages/examples/usage-tracking/usage-tracking.js
benlower: i found a js version of tiktoken (js-tiktoken) and have moved to that so we get consistent numbers across the js and python versions of this script.

packages/examples/usage-tracking/usage-tracking.js
benlower: yes it is. changed the implementation to handle paging.

PR #237, opened 2024-03-08
Mute agent when not SPEAKING and support TTS transcripts delivered as deltas

mdepinet:


packages/fixie-web/src/voice.ts
```diff
@@ -339,7 +341,7 @@ export class VoiceSession {
     } else if (msg.type === 'output') {
       this.handleOutputChange(msg.text, msg.final);
     } else if (msg.type === 'voice_synced_transcript') {
-      this.handleOutputTranscript(msg.text);
```
juberti, line 342: should we just change this to make onOutputChange always expose the synced transcript, rather than having separate callbacks for slightly different transcripts?

I'd also prefer to retain the "final" argument to onOutputChange, which seems straightfoward to accommodate.

packages/fixie-web/src/voice.ts
```diff
@@ -212,6 +213,7 @@ export class VoiceSession {
     if (state != this._state) {
       console.log(`[voiceSession] ${this._state} -> ${state}`);
       this._state = state;
+      this.audioElement.muted = state != VoiceSessionState.SPEAKING;
```
juberti, line 216: I think you're going to need some fudge factor here to prevent clipping the end of speech. Probably what this means is that we're going to need a different state to signal interruptions vs regular speaking turn change (which will need a more graceful transition).

packages/fixie-web/src/voice.ts
mdepinet: I think it makes sense to remove callbacks related to LLM generation.  We can't have a "final" argument in the TTS transcript case until I implement that server-side, so my proposal would be to ship this, update the server to send {delta, final} instead of {text}, then update this to combine all the output-related callbacks into one with the full tts-synced transcript text and whether it's final.

Tangentially related, I'd love to pull this SDK into our monorepo once we're the only users.

packages/fixie-web/src/voice.ts
mdepinet: Long term you may be right, but this seems to do the trick for now.

juberti: Well... if it's in the client it's harder to tweak later. Plus this changes it for everyone regardless of enable_interruptions. 

Maybe include a separate argument regarding whether to stop play out that we can tweak independently from the server side?

juberti: Then we can only set the stop-playout flag on an interruption.

juberti: From the comments in the other issue, it sounds like clipping will be a corner case, so I'm OK to revisit this later/if-needed.

packages/fixie-web/src/voice.ts
```diff
@@ -76,6 +76,7 @@ export class VoiceSession {
   private inAnalyzer?: StreamAnalyzer;
   private outAnalyzer?: StreamAnalyzer;
   private pinger?: ReturnType<typeof setInterval>;
+  private _outputTranscript: string = '';
```
juberti, line 79: TS style is to not use _ on private ivars.

packages/fixie-web/src/voice.ts
juberti: That plan sounds good.

juberti (approved):
LG to proceed with the plan mentioned here

PR #238, opened 2024-03-11
Update tts transcript handling for server impl. Also remove old output stuff and replace with the tts-synced version.

mdepinet:
In light of our Friday conversation, I'm planning on breaking the client briefly while the server and client changes roll out out-of-sync.  (Will have a client PR to use this SDK version after standup)

### DIFF
--- a/packages/fixie-web/src/voice.ts
+++ b/packages/fixie-web/src/voice.ts
@@ -84,11 +84,8 @@ export class VoiceSession {
   /** Called when the input changes. */
   onInputChange?: (text: string, final: boolean) => void;
 
-  /** Called when the output changes. Note that this may fire *before* speech has completed playing. */
-  onOutputChange?: (text: string, final: boolean) => void;
-
   /** Called as output voice is played with the text that should have completed playing (for this response). */
-  onOutputTranscript?: (text: string) => void;
+  onOutputChange?: (text: string, final: boolean) => void;
 
   /** Called when new latency data is available. */
   onLatencyChange?: (metric: string, value: number) => void;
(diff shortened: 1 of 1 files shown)


packages/fixie-web/src/voice.ts
```diff
@@ -356,20 +351,15 @@ export class VoiceSession {
...
+    } else if (msg.delta) {
       this.outputTranscript += msg.delta;
-      this.onOutputTranscript?.(this.outputTranscript);
-      if (msg.final) {
-        this.outputTranscript = '';
-      }
-    } else {
-      this.onOutputTranscript?.(msg.text);
+    }
+    this.onOutputChange?.(this.outputTranscript, msg.final);
```
juberti, line 360: In the case where only `text` is filled in, are we OK with this being undefined?

juberti (approved):

packages/fixie-web/src/voice.ts
mdepinet: I don't understand your question.  What is `this`?  With the exception of onOutputChange, nothing here should be undefined once the server change lands.

mdepinet: Perhaps you're referring to the interim case before the server change lands. In that case, AI.Town doesn't use onOutputChange right now, so it won't make any difference.  (It uses onOutputTranscript, which means it won't show agent transcripts until the server piece lands.)

mdepinet: I'm going to start merging things so I can get everything landed today. If there are issues, then I'll fix them later. :)

